PR #73, opened 2021-07-19
Add radio buttons

EllenWiderstrand:


src/components/controls/radios/DefaultRadio/DefaultRadio.tsx
```diff
@@ -0,0 +1,38 @@
...
+    background: ${({ theme }) => theme.colors.background.secondary};
+  }
+`;
+
+export interface DefaultRadioProps
+  extends InputHTMLAttributes<HTMLInputElement> {
+  onChange: (e: ChangeEvent<HTMLInputElement>) => void;
+}
+
+export const DefaultRadio = ({ ...props }: DefaultRadioProps) => {
```
filiptammergard, line 36: I wonder if this DefaultRadio is needed. Would it be possible to just have a Radio, which result in a DefaultRadio if label is not passed, and LabelRadio if label is passed. What do you think?

filiptammergard, line 36: Instead of destructuring and spreading you can do just `props` 😊

EllenWiderstrand: I can try that out! 😊

filiptammergard (approved):
👍

PR #77, opened 2021-07-29
Small adjustments to form elements to suit how they are used 

EllenWiderstrand:


src/components/controls/selects/DefaultDropdownSelect/DefaultDropdownSelect.tsx
```diff
@@ -56,8 +61,8 @@ export const DefaultDropdownSelect = ({
         </option>
       )}
       {options.map((option) => (
-        <option key={option} value={option}>
-          {option}
+        <option key={option.key ?? option.value} value={option.value}>
```
filiptammergard, line 64: Should key perhaps be required?

filiptammergard (approved):

src/components/controls/selects/DefaultDropdownSelect/DefaultDropdownSelect.tsx
EllenWiderstrand: I was considering that, but e.g. the website only uses value and I assume that value (almost always) will be unique. Do you think it is better to make it required?

filiptammergard: That's true, I think this is fine! 👌 As you said, options shouldn't be the same, and in a corner case where options are the same, keys can be passed.

PR #95, opened 2021-08-24
feat: add customtable

bjornlexell:
CustomTable is a debatable name :D

### DIFF
--- a/src/components/table/CustomTable/CustomTable.tsx
+++ b/src/components/table/CustomTable/CustomTable.tsx
@@ -1,0 +1,24 @@
+import { CSSProperties, ReactNode } from "react";
+import { TableRoot } from "../TableRoot";
+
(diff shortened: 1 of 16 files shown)


filiptammergard:
@bjornlexell I realize we need to enforce having `React` in scope again, to support React 16. So I added this: https://github.com/einride/ui/pull/99

Let's merge https://github.com/einride/ui/pull/99, then resolve conflicts here and ensure linting is ok, and I'll review again. :)

bjornlexell:
> @bjornlexell I realize we need to enforce having `React` in scope again, to support React 16. So I added this: #99
> 
> Let's merge #99, then resolve conflicts here and ensure linting is ok, and I'll review again. :)

Sounds like a great idea to add an error for that rule (obv... since I forgot it in the first PR after adding support for reat 16)

PR #239, opened 2022-03-04
feat: add site with charger pin component

jenseo:


src/components/mapping/SiteWithChargerPin/SiteWithChargerPin.tsx
```diff
@@ -0,0 +1,65 @@
...
+  )
+})
+
+type Size = "sm" | "md"
+
+const StyledSitePin = styled(SitePin)<{ size: Size }>`
+  position: absolute;
+  top: 0;
+  left: 0;
+  z-index: 1;
```
filiptammergard, line 29: We have been trying to avoid using `z-index` in the component library, since it potentially could mess up in the applications using a map. What happens if we remove those? Is there another solution?

jenseo: Makes sense, I can let them stack in order of apperance instead, z-index was just a way of playing it safe. Then I believe it would be safe also in a map context with other z-indexes potentially flying around

PR #241, opened 2022-03-04
feat: add map pod component

jenseo:


src/components/mapping/MapPod/MapPod.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { PodIcon } from "./PodIcon"
+
+export type MapPodProps = SVGAttributes<SVGSVGElement>
+
+export const MapPod = forwardRef<SVGSVGElement, MapPodProps>((props, ref) => {
+  return <StyledPodIcon {...props} ref={ref} />
+})
+
+const StyledPodIcon = styled(PodIcon)`
+  filter: drop-shadow(0 2px 2px rgba(18, 18, 18, 0.32));
```
filiptammergard, line 12: What does this do? 🙂

src/components/mapping/MapPod/PodIcon.tsx
```diff
@@ -0,0 +1,59 @@
+import { forwardRef, SVGAttributes } from "react"
+
+export type PodIconProps = SVGAttributes<SVGSVGElement>
+
+export const PodIcon = forwardRef<SVGSVGElement, PodIconProps>(
```
filiptammergard, line 5: Why do you think this should be split into two components? 🙂

src/components/mapping/MapPod/MapPod.tsx
jenseo: It's a filter for applying the shadow of the icon seen in Figma. Hard to export it with the icon and keep the size of it.

src/components/mapping/MapPod/PodIcon.tsx
jenseo: I wanted to apply the shadow to it and wasn't sure if we wanted to do it as a styled component or not :) But an easy change if we should do it another way 👍

src/components/mapping/MapPod/MapPod.tsx
jenseo: We could add a normal drop shadow of course, since it's a rather uncomplicated svg without transparency. So a filter might be bit over the top :)

filiptammergard (approved):

PR #281, opened 2022-03-31
fix: add width to linear progress wrapper

jenseo:
making sure the child of linear progress always have space to calculate percentage from

### DIFF
--- a/src/components/charts/LinearProgress/LinearProgress.tsx
+++ b/src/components/charts/LinearProgress/LinearProgress.tsx
@@ -47,6 +47,8 @@ const Wrapper = styled.div`
   height: ${({ theme }) => theme.spacer}px;
   border-radius: ${({ theme }) => theme.spacer}px;
   position: relative;
+  /* Width needed to make sure component takes up full width in flex containers */
+  width: 100%;
 `
 
 const Value = styled.div<{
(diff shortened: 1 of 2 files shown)


src/components/charts/LinearProgress/LinearProgress.tsx
```diff
@@ -47,6 +47,8 @@ const Wrapper = styled.div`
   height: ${({ theme }) => theme.spacer}px;
   border-radius: ${({ theme }) => theme.spacer}px;
   position: relative;
+  /* Width needed to make sure the child has space to use */
```
filiptammergard, line 50: Hmm, isn't the case more like "Width needed to make sure component takes up full width in flex containers"? 🙂

filiptammergard (changes requested):
Can you add this for StepProgress as well? Should have the same problem!

src/components/charts/LinearProgress/LinearProgress.tsx
jenseo: Yeah, that might be a better description perhaps! Thanks, will change!

jenseo:
> 

That's true, I actually thought it would take up the whole space since the Wrapper in the step progress is display flex, but that was a snedtänk from my side. Tried it out and you're definitely correct :)

filiptammergard (approved):

PR #282, opened 2022-03-31
feat: make inputs without labels rounded

filiptammergard:
Decided together with Rui that all input components without labels should have rounded corders. This should make it possible for you to use `DefaultDropdownSelect` out of the box in Explore @jenseo!


jenseo (approved):
Awesome!!

PR #287, opened 2022-04-05
add map icons as importable svg's

jenseo:


src/components/mapping/SvgIcons/SvgIcons.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <code
+      style={{
+        display: "block",
+        marginTop: "15px",
+        padding: "4px",
+        backgroundColor: "#fafafa",
+      }}
+    >
+      {`import ${filename} from "@einride/ui/lib/mapIcons/${filename}.svg"`}
+    </code>
```
filiptammergard, line 26: This is clever, but I think we should continue only rendering the actual component on the canvas. Instead we can add something like this below `title` in the default export, that gets added to the docs page:

```ts
parameters: {
    docs: {
      description: {
        component:
          'Import Charger Small like this: `import chargerSmall from "@einride/ui/lib/mapIcons/chargerSmall.svg"`',
      },
    },
  },
```

And since these are separate "components" if think we should continue having separating these into the respective story files as before. That would make it possible to add proper alt text and provide clear import instructions in the docs.

What do you think?

src/components/mapping/SvgIcons/SvgIcons.stories.tsx
```diff
@@ -0,0 +1,68 @@
+import { Story } from "@storybook/react/types-6-0"
+
+export default {
+  title: "Mapping/SvgIcons",
+  argTypes: {
+    filename: {
+      table: {
+        disable: true,
+      },
+    },
```
filiptammergard, line 10: What does this do?

src/components/mapping/SvgIcons/SvgIcons.stories.tsx
jenseo: Absolutely, sounds great! Will implement this!

src/components/mapping/SvgIcons/SvgIcons.stories.tsx
jenseo: It makes sure the filename prop line isn't added to the controls list. But if we separate them into different components, this will go away :)

PR #297, opened 2022-04-12
fix: add topGap on linearGauge component

jenseo:


src/components/charts/LinearGauge/LinearGaugeProgress.tsx
```diff
@@ -15,18 +15,24 @@ export const LinearGaugeProgress = ({
   responsiveRadius,
   strokeWidth,
 }: LinearGaugeProgressProps): JSX.Element => {
+  const topGap = -4
+  const circleStartPoint = (360 * topGap) / 200 + -90
```
filiptammergard, line 19: I think a good practice for constants like this is to use upper snake case and put them outside of the components. Also the suffix them with the unit. Could be like this:

```tsx
const TOP_GAP_PIXELS = -4
const CIRCLE_START_POINT_DEGREES = (360 * topGap) / 200 + -90
```

Then you don't have to pass them as props bu can use them directly.

But it's a bit difficult to understand what the second constant is doing based on the name. Is it a kind of offset? `200 + -90` looks weird, why not `200 - 90`?

filiptammergard (changes requested):

src/components/charts/LinearGauge/LinearGaugeProgress.tsx
jenseo: Good idea, will make them constants outside instead. 

For the other question, I had the degree as a fixed negative value at first, that's why I had to do + -. Forgot to change that. Pushing an update!

src/components/charts/LinearGauge/LinearGaugeProgress.tsx
```diff
@@ -37,24 +44,34 @@ export const LinearGaugeProgress = ({
...
   percentage: number
   strokeWidth: number
+  topGap: number
+  circleStartPoint: number
 }>`
   fill: none;
-  stroke-dasharray: ${({ percentage }) => `${percentage},${100}`};
+  stroke-dasharray: ${({ percentage, topGap }) =>
+    `${percentage + topGap / 2 > 0 ? percentage + topGap / 2 : 0},${100}`};
+  stroke-dashoffset: ${({ topGap }) => `${topGap}`};
```
filiptammergard, line 69: Now you don't need to send these as props anymore, but can use them directly 🙂 

```suggestion
  stroke-dashoffset: ${TOP_GAP};
```

filiptammergard (changes requested):

PR #299, opened 2022-04-12
chore: add contribute to readme

stenehall:


stenehall:
> Thanks!

Just make sure all I wrote is correct :) Since I don't 100% know what flow you want :P

PR #300, opened 2022-04-12
feat: add title to popover

stenehall:
This adds two things, the possibility to provide a title to the popover as a "full width" scaling for the responsive modes (no idea if this is an acceptable change from design or not)

popover_before mov
popover_after mov


src/components/views/Popover/Popover.tsx
```diff
@@ -40,33 +45,35 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
...
-                  )}
-                  {primaryAction && (
-                    <PrimaryButton onClick={primaryAction.handler}>
-                      {primaryAction.text}
-                    </PrimaryButton>
-                  )}
-                </Actions>
-              </MediumLargeNav>
+              <Header>
+                <Title3>{title}</Title3>
```
stenehall, line 61: This will generate an empty tag if no title is provided, is that ok? Or do we want something like `{title ? <Title3>{title}</Title3> : <span />}`

filiptammergard: I updated to this instead!

filiptammergard (approved):

PR #301, opened 2022-04-12
feat: add StepVerticalBar progress chart

jenseo:


src/components/charts/StepVerticalBar/StepVerticalBar.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </Wrapper>
+    )
+  },
+)
+
+const Wrapper = styled.div`
+  display: flex;
+  flex-direction: column-reverse;
+  gap: ${({ theme }) => theme.spacer}px;
+  width: 16px;
```
filiptammergard, line 49:
```suggestion
  width: ${({ theme }) => 2 * theme.spacer}px;
```

src/components/charts/StepVerticalBar/StepVerticalBar.tsx
```diff
@@ -0,0 +1,58 @@
+import styled from "@emotion/styled"
+import { forwardRef, HTMLAttributes } from "react"
+import { ContentColor } from "../../../lib/theme/types"
+
+interface StepVerticalBarBaseProps extends HTMLAttributes<HTMLDivElement> {
```
filiptammergard, line 5: I'm having a discussion with Rui right now what this component should be called. Since it's kind of exactly the same as `<StepProgress>` component except it's vertical instead of horizontal, I think it'd make sense to call it `<StepVerticalProgress>`. But I'll get back to you when we have decided!

filiptammergard (commented):
Good job, looks great. Let's hold until I know what it should be called. I'll let you know!

src/components/charts/StepVerticalBar/StepVerticalBar.tsx
filiptammergard: I talked to Rui about this, let's call this one `<StepVerticalProgress>` as discussed 🙂

filiptammergard (approved):

PR #307, opened 2022-04-22
chore: remove useDisclosure that now lines in @einride/hooks

filiptammergard:
All repos have already migrated to @einride/hooks, so no need to make this a breaking change.


jenseo (approved):
⭐

PR #322, opened 2022-06-15
feat: respect reduce motion preference for framer animations by default

filiptammergard:
It's still fully configurable for the consumer of EinrideProvider.
This will switch off transform animations in Framer Motion animations by default.


jenseo (approved):
I think this is great, making it possible to control it if needed 👍

src/contexts/EinrideProvider.tsx
```diff
@@ -12,6 +13,13 @@ interface EinrideProviderProps {
   colorMode?: ColorMode
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   customTheme?: any
+  /**
+   * If set to "user", Framer Motion will respect the system "Reduce motion" setting by switching
+   * transform animations used in Framer Motion animations off automatically.
+   *
+   * Default is "user".
+   */
+  framerReducedMotion?: "always" | "never" | "user"
```
bjornlexell, line 22: Since this is a context you could probably just skip prop this and use the context with a more specific parameter when you want a different option. 

You could also use this approach deeper down in the tree aswell

```suggestion
  <EinrideProvide> // turns it off :(
      <MotionConfig framerReducedMotion="always"> // its back! :)
        <MyApp />
      </MotionConfig
  </EinrideProvider>
```
Just another way of doing it =)

bjornlexell (approved):
LGTM :gem:

PR #324, opened 2022-06-15
feat(avatar): support size and radius options

filiptammergard:
- feat(theme): update border-radii theme values
- feat(avatar): support size and radius options


jenseo (commented):
🌟

jenseo (approved):

PR #334, opened 2022-06-17
feat: add sheets component

filiptammergard:
This component will replace ElevatedSheets and PlainSheets, with prop that controls whether overlay should be used.


jenseo (approved):
One component to rule them all. Like it!

PR #338, opened 2022-06-22
fix: remove webpack-specific %PUBLIC_URL% variable

jenseo:


.storybook/manager-head.html
```diff
@@ -1 +1 @@
-<link rel="icon" href="%PUBLIC_URL%/favicon.svg" />
+<link rel="icon" href="/favicon.svg" />
```
filiptammergard, line 1: image

:D

filiptammergard (approved):
Nice!

.storybook/manager-head.html
jenseo: OMG!!

PR #345, opened 2022-07-04
fix: make sure table row separator works in safari

jenseo:


src/components/table/Thead/Thead.tsx
```diff
@@ -17,7 +17,7 @@ export const Thead = forwardRef<HTMLTableSectionElement, TheadProps>(
 )
 
 const StyledThead = styled.thead`
-  tr {
+  tr:before {
```
filiptammergard, line 20: Does `thead` get a box-shadow in Safari unless `:before` is unsed?

src/components/table/Tr/Tr.tsx
```diff
@@ -15,5 +15,14 @@ export const Tr = forwardRef<HTMLTableRowElement, TrProps>(({ children, ...props
 })
 
 const StyledTr = styled.tr`
-  box-shadow: 0 -1px 0 ${({ theme }) => theme.colors.border.primary};
+  // display: block needed for box-shadow in safari, can't be set on table row directly
```
filiptammergard, line 18: A bit unclear to me, since `:before` is an inline element by default right? Maybe something like this?

> // box-shadow on table element is not supported in Safari

src/components/table/Tr/Tr.tsx
```diff
@@ -15,5 +15,14 @@ export const Tr = forwardRef<HTMLTableRowElement, TrProps>(({ children, ...props
...
+  &:before {
```
filiptammergard, line 19: Two colons seems to be the preferred way: https://developer.mozilla.org/en-US/docs/Web/CSS/::before

```suggestion
  &::before {
```

filiptammergard (approved):
Cool!

src/components/table/Tr/Tr.tsx
jenseo: Good suggestion!

src/components/table/Thead/Thead.tsx
jenseo: The `tr` child of `thead` gets a box-shadow in both Safari and Chrome if we don't reset it, since it's applied on `::before` on the "normal" `tr::before`'s

PR #349, opened 2022-07-06
feat: add label option to SearchSelectInput component

jenseo:


src/components/controls/selects/SearchSelect/SearchSelect.tsx
```diff
@@ -201,9 +217,10 @@ const Wrapper = styled.div<{ isFullWidth?: boolean }>`
 
 const OptionsWrapper = styled.div`
   position: absolute;
-  top: ${({ theme }) => 6 * theme.spacer}px;
+  top: ${({ theme }) => `calc(100 % + ${theme.spacer * 2})px`};
```
filiptammergard, line 220:
```suggestion
  top: ${({ theme }) => `calc(100% + ${theme.spacer * 2})px`};
```

src/components/controls/selects/SearchSelect/SearchSelect.tsx
```diff
@@ -201,9 +217,10 @@ const Wrapper = styled.div<{ isFullWidth?: boolean }>`
...
   right: 0;
   left: 0;
+  width: 100%;
```
filiptammergard, line 223: Either `right` and `left` or `width` should do it. I think it makes sense to use `right` and `left` for absolute positioned elements, so I suggest going for that!

filiptammergard (approved):
I noticed this happens when you pass a `message` with the new positioning of the options wrapper.

image

But not seeing any apparent solution for it now...

src/components/controls/selects/SearchSelect/SearchSelect.tsx
jenseo: Sure!

jenseo:
> I noticed this happens when you pass a `message` with the new positioning of the options wrapper.
> 
> image
> 
> But not seeing any apparent solution for it now...

Hmm, yeah, I'll take a thinking round on that one!

PR #357, opened 2022-07-11
fix: allow primary actions on popover buttons

jenseo:
To enable using props like this:

https://github.com/einride/saga-admin/blob/146fe5955ded4060e7692c2227f3329ce2448897/src/components/OrganizationUsersList.tsx#L100


filiptammergard (approved):
🙏

PR #358, opened 2022-07-13
chore(theme): build theme type from explicit types

filiptammergard:
Prevents accidental theme form updates.


jenseo (approved):
Seems some imports need to be updated here and there, apart from that it looks awesome!

PR #359, opened 2022-07-13
fix: improve slider to make it useable

jenseo:


src/components/controls/sliders/Slider/Slider.tsx
```diff
@@ -1,37 +1,85 @@
...
-  ({ onInputChange, value = 0, ...props }, ref) => {
+  ({ max = 100, min = 0, onInputChange, onInputChangeEnd, value = 0, ...props }, ref) => {
     const [hasFocus, setHasFocus] = useState(false)
+    const [internalValue, setInternalValue] = useState(value)
+    // If we work with values other than 0-100,
+    // we need to calculate the percentage to get the correct position of the slider items
+    const shouldCalculatePercentage = useMemo(() => max !== 100 || min !== 0, [max, min])
+
+    const currentValue = (): number => {
+      const valueToEvaulate = value !== 0 ? value : internalValue
```
filiptammergard, line 36: Why can't we always use `internalValue` for this?

src/components/controls/sliders/Slider/Slider.tsx
```diff
@@ -1,37 +1,85 @@
...
 }
 
 export const Slider = forwardRef<HTMLInputElement, SliderProps>(
-  ({ onInputChange, value = 0, ...props }, ref) => {
+  ({ max = 100, min = 0, onInputChange, onInputChangeEnd, value = 0, ...props }, ref) => {
     const [hasFocus, setHasFocus] = useState(false)
+    const [internalValue, setInternalValue] = useState(value)
+    // If we work with values other than 0-100,
+    // we need to calculate the percentage to get the correct position of the slider items
+    const shouldCalculatePercentage = useMemo(() => max !== 100 || min !== 0, [max, min])
```
filiptammergard, line 33: Calculating a percentage is such a quick operation in JS, so I think we can remove this and always use the same `getCurrentValue` function for this. What do you think about that?

jenseo: Sure!

src/components/controls/sliders/Slider/Slider.tsx
jenseo: I guess we would need a useEffect then if we want to change the internalValue with value, without dragging the slider

filiptammergard: What if you're actually passing value = 0 from the outside? Could that produce something weird?

jenseo: It wouldn't update internalValue without a useEffect. But only thing it would do is "drag" the slider to zero. (If zero is minimum)

jenseo: If passed after initialization I mean :)

filiptammergard: Sounds good! Gonna test this in Storybook when time permits, but go for it. 🙂

jenseo: The way I though about it:

internalValue would be used if the component is not used as a `controlled`component. Value would be used when we want to control the component more from the outside

jenseo: But you're right, if the user controls it from the outside and pass value === 0 after initialization, it would probably break also in this implementation. Will give it a thought!

filiptammergard (approved):
Sweet!

PR #384, opened 2022-07-28
fix: dependabot correct name

Jontii:


Jontii:
Linkin this here for the issue https://github.com/chromaui/chromatic-cli/issues/550

filiptammergard (approved):

PR #424, opened 2022-08-12
fix: correctly show disabled styles on controls

filiptammergard:


radhus (approved):
I don't know CSS that good so I mostly trust your changes here. But I was expecting to see any change involving a `color:` key, to get the lighter color on the input text. Is that inferred by some other change here?

src/components/controls/inputs/BaseInput/BaseInput.tsx
```diff
@@ -131,8 +131,10 @@ const StyledInput = styled.input<{ hasLabel: boolean; leftIcon: boolean; rightIc
     color: ${({ theme }) => theme.colors.content.secondary};
   }
 
+  &:disabled,
   &:disabled::placeholder {
     color: ${({ theme }) => theme.colors.content.tertiary};
+    cursor: not-allowed;
   }
```
filiptammergard, line 138: This is what fixed the specific bug. Before, we only styled the placeholder text with the disabled text color. Now, we're styling. both the placeholder text and the value text. We're also adding the `not-allowed` cursor styling to indicate clearly the field is disabled.

@radhus

radhus: Ah, nice! Thanks for the explanation!

PR #490, opened 2022-09-09
fix(map-icons): scale up svgs to reduce blurriness in deckgl maps

filiptammergard:
Deckgl renders SVGs in a weird way that makes icons blurry... Scaling up solves it (tried out in Evolve).


jenseo (approved):
This is great!

PR #529, opened 2022-09-20
chore: do not pass color prop to component

filiptammergard:
Right now, the `color` prop is leaking to the DOM, which is shouldn't do.

image

Using `shouldForwardProp` fixes that.


jenseo (approved):
Cool!

PR #627, opened 2022-10-26
feat(Stack): add flexible stack component

filiptammergard:


src/components/layout/Stack/Stack.tsx
```diff
@@ -0,0 +1,60 @@
...
+interface WrapperProps {
+  align: AlignItems
+  gap: Gap
+  justify?: JustifyContent
+  width?: Width
+}
+
+const Wrapper = styled.div<WrapperProps>`
+  display: flex;
+  flex-direction: column;
```
jenseo, line 42: Without knowing your reasoning and plans going forward, would it make sense to also make it possible to send in direction horizontal/vertical with `column` or `row` for instance?

jenseo (approved):

PR #628, opened 2022-10-26
feat(Text): add text component

filiptammergard:


jenseo (approved):

src/lib/theme/types.ts
```diff
@@ -38,32 +38,37 @@ interface ButtonIcon {
...
-  | "thin"
-  | "light"
-  | "normal"
-  | "book"
-  | "medium"
-  | "semibold"
-  | "bold"
-  | "extrabold"
-  | "black"
+const fontWeights = [
```
mblodau1337, line 59: Not 100% sure but just noticing a divergence from a similar pattern further up in the file where you did an `export const contentColors`. This row doesn't have an export. Working as intended?

mblodau1337 (approved):
Looking good! Left a comment about a divergence from a pattern further up. No idea though if it's something to be fixed or working as intended.

src/lib/theme/types.ts
filiptammergard: Nice catch! `FontWeight` is not needed outside of this module (yet), so leaving it like this for now 🙂

PR #640, opened 2022-10-28
feat: export z-index object to make it easier to work with z-indices

filiptammergard:


jenseo (approved):
This will be awesome!

PR #881, opened 2023-01-10
fix(DatePicker): use luxon to create expected date strings

LucidityDesign:


package.json
```diff
@@ -32,7 +32,8 @@
     "@radix-ui/react-tabs": "1.0.1",
     "@radix-ui/react-tooltip": "1.0.2",
     "dayjs": "1.11.7",
-    "lodash.merge": "4.6.2"
+    "lodash.merge": "4.6.2",
+    "luxon": "3.2.1"
```
filiptammergard, line 36: Let's put this as a dev dependency, so that it is not included to the library consumers 🙂

src/components/controls/dates/DatePicker/DatePicker.stories.tsx
```diff
@@ -95,17 +94,20 @@ Mouse.args = {
...
+  date = date.set({ day: 1 })
+  await expect(input).toHaveValue(date.toFormat(DATE_FORMAT))
   await userEvent.click(input)
   const previousMonthButton = canvas.getAllByRole("button")[0]
   await userEvent.click(previousMonthButton)
   const firstDayInLastMonthButton = canvas.getByRole("button", { name: "1" })
   await userEvent.click(firstDayInLastMonthButton)
-  await expect(input).toHaveValue(`${new Date().getFullYear()}-${new Date().getMonth()}-01`)
+  date = date.minus({ month: 1 })
+  await expect(input).toHaveValue(date.toFormat(DATE_FORMAT))
```
filiptammergard, line 110: And like this, and similar in the other places?

```suggestion
  const firstDayInPreviousMonth = firstDayInCurrentMonth.minus({ month: 1 })
  await expect(input).toHaveValue(firstDayInPreviousMonth.toFormat(DATE_FORMAT))
```

filiptammergard (approved):
👏

PR #887, opened 2023-01-11
Calendar

LucidityDesign:
- feat(Calendar): add standalone calendar component

### DIFF
--- a/src/components/controls/dates/Calendar/Calendar.stories.tsx
+++ b/src/components/controls/dates/Calendar/Calendar.stories.tsx
@@ -1,0 +1,83 @@
+import { expect } from "@storybook/jest"
+import { ComponentMeta, ComponentStory } from "@storybook/react"
+import { userEvent, within } from "@storybook/testing-library"
+import { useState } from "react"
+import { Calendar } from "./Calendar"
+
+export default {
+  title: "Controls/Dates/Calendar",
+  component: Calendar,
+} as ComponentMeta<typeof Calendar>
+
+const Template: ComponentStory<typeof Calendar> = (args) => {
+  const { value: defaultValue } = args
+  const [value, setValue] = useState(defaultValue)
+  return <Calendar {...args} value={value} onChange={setValue} />
+}
+
+export const Default = Template.bind({})
+Default.args = {}
+Default.play = async ({ args }) => {
+  await expect(args.value).toBeFalsy()
+}
+
+export const DefaultValue = Template.bind({})
+DefaultValue.args = {
+  value: new Date(new Date().setDate(10)),
+}
+DefaultValue.play = async ({ canvasElement, args }) => {
+  const canvas = within(canvasElement)
+  await expect(args.value).toBeTruthy()
+
+  const currentDayButton = canvas.getByRole("button", { name: "10" })
+  await expect(currentDayButton.getAttribute("data-selected")).toBe("true")
+}
+
+export const Mouse = Template.bind({})
+Mouse.args = {
+  value: new Date(new Date().setDate(10)),
+}
+Mouse.play = async ({ canvasElement, args }) => {
+  const canvas = within(canvasElement)
+  await expect(args.value).toBeTruthy()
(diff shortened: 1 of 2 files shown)


src/components/controls/dates/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { Calendar as MantineCalendar } from "@mantine/dates"
+import { useTheme } from "../../../../hooks/useTheme"
+import { Box, BoxProps } from "../../../layout/Box/Box"
+
+interface CalendarBaseProps {
+  /** Maximum possible date. */
+  maxDate?: Date
+
+  /** Minimum possible date. */
+  minDate?: Date
```
LucidityDesign, line 11: Not sure if we need `minDate`. We don't have it in DatePicker

src/components/controls/dates/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  /** Called when date changes. */
+  onChange: (value: CalendarValue) => void
```
LucidityDesign, line 14: As the calendar is always controlled, I set `onChange` as required.

src/components/controls/dates/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  /** Selected date, required with controlled input. */
+  value?: CalendarValue
+
+  /** Props passed to root element. */
+  wrapperProps?: BoxProps
+
+  /** Accessible name */
+  label: string
```
LucidityDesign, line 23: I'm not sure how you handle a11y. If the component is used with a descriptive headline we don't need a aria-label and I should change it to optional

src/components/controls/dates/Calendar/Calendar.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const currentDayButton = canvas.getByRole("button", { name: "10" })
+  expect(currentDayButton.getAttribute("data-selected")).toBe("true")
+}
+
+export const Mouse = Template.bind({})
+Mouse.args = {
+  value: new Date(new Date().setDate(10)),
+}
+
+Mouse.play = async ({ canvasElement, args }) => {
```
filiptammergard, line 48: I have sticked to grouping the template, args and interaction tests without empty lines between to make it easy to see the full "story". What do you think about doing that here as well?

```suggestion
export const Mouse = Template.bind({})
Mouse.args = {
  value: new Date(new Date().setDate(10)),
}
Mouse.play = async ({ canvasElement, args }) => {
```

filiptammergard (commented):
Great! I think we can hold off merge here until tomorrow so that we can show it for Rui and ask for feedback.

src/components/controls/dates/Calendar/Calendar.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Calendar } from "./Calendar"
+
+export default {
+  title: "Controls/Dates/Calendar",
+  component: Calendar,
+  argTypes: {
+    disabled: {
+      control: "boolean",
+    },
+  },
```
filiptammergard, line 14: I think this should be removed, as we don't have a `disabled` prop, right?

src/components/controls/dates/Calendar/Calendar.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+} as ComponentMeta<typeof Calendar>
+
+const Template: ComponentStory<typeof Calendar> = (args) => {
+  const { value: defaultValue } = args
+  const [value, setValue] = useState<Date | null>(defaultValue || null)
```
filiptammergard, line 19: What do you think about inferring types like this?

```suggestion
  const [value, setValue] = useState(defaultValue)
```

filiptammergard (approved):
Nice! When these comments are resolved I think this is ready to go 🙂

LucidityDesign:
Thanks for the great Feedback

PR #888, opened 2023-01-11
chore(DateRangePicker): use 10th day instead of today

LucidityDesign:


filiptammergard (approved):
Why was this needed? 🙂

LucidityDesign:
The expected was always the 10th, the received date was "today". That's why it succeeded yesterday (10th of January) and fails today

filiptammergard:
> The expected/default value was always the 10th, the received date was "today". That's why it succeeded yesterday (10th of January) and fails today See here: https://github.com/einride/ui/actions/runs/3892029088/jobs/6642961416

I see, looks good! Maybe Luxon can be used where `defaultValue` is set as well for consistency? But that can be done later 👍 Merge away!

PR #896, opened 2023-01-12
fix(Sheets): prevent content underneath actions

LucidityDesign:


src/components/views/Sheets/Sheets.tsx
```diff
@@ -169,6 +171,7 @@ const DialogContent = styled(motion(Dialog.Content))<{ size: Size }>`
     inset-inline-start: unset;
     border-end-start-radius: ${({ theme }) => theme.borderRadii.lg};
     border-end-end-radius: ${({ theme }) => theme.borderRadii.lg};
+    display: block;
```
LucidityDesign, line 174: This is not needed. However I tried to keep the original display of the component

src/components/views/Sheets/Sheets.tsx
```diff
@@ -230,9 +225,8 @@ const Content = styled.div`
 `
 
 const SmActions = styled.nav`
-  position: fixed;
-  inset-inline: ${({ theme }) => 2 * theme.spacingBase}rem;
-  inset-block-end: ${({ theme }) => 3 * theme.spacingBase}rem;
+  margin-inline: ${({ theme }) => 2 * theme.spacingBase}rem;
+  margin-block: ${({ theme }) => theme.spacingBase}rem ${({ theme }) => 2 * theme.spacingBase}rem;
```
filiptammergard, line 229: I would do it like this here which I think is a bit easier to read, but you can choose!

```suggestion
  margin-block-start: ${({ theme }) => theme.spacingBase}rem;
  margin-block-end: ${({ theme }) => 2 * theme.spacingBase}rem;
```

filiptammergard (changes requested):
In this case, the commit message should be a `fix`, to make sure a patch version is released when merging. The other two suggestions are optional. Looks good!

filiptammergard (approved):
Nice!

PR #897, opened 2023-01-12
feat(Tooltip): enable width prop

filiptammergard:
Makes multiline tooltips possible:
image

### DIFF
--- a/src/components/content/Tooltip/Tooltip.stories.tsx
+++ b/src/components/content/Tooltip/Tooltip.stories.tsx
@@ -101,6 +101,19 @@ Table.play = async ({ canvasElement }) => {
   await expect(tooltipTrigger).toHaveAttribute("data-state", "closed")
 }
 
+export const Multiline = Template.bind({})
+Multiline.args = {
+  children: "Tooltip trigger",
+  content:
+    "Some really really really really really really really really really really really really really really really really really really really really really really really long tooltip content.",
+  width: 50,
+}
+Multiline.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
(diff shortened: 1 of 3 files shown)


src/components/content/Tooltip/Tooltip.tsx
```diff
@@ -20,6 +22,9 @@ interface TooltipProps {
 
   /** Merges the original component props with the props of the supplied component and change the underlying DOM node. */
   triggerAsChild?: boolean
+
+  /** Width of the tooltip. */
+  width?: Width
```
LucidityDesign, line 27: can you remove that prop from the HTML attributes, please?

src/components/content/Tooltip/Tooltip.stories.tsx
```diff
@@ -101,6 +101,19 @@ Table.play = async ({ canvasElement }) => {
...
+export const Multiline = Template.bind({})
+Multiline.args = {
+  children: "Tooltip trigger",
+  content:
+    "Some really really really really really really really really really really really really really really really really really really really really really really really long tooltip content.",
+  width: 50,
+}
+Multiline.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const tooltipTrigger = canvas.getByRole("button", { name: "CO2e" })
```
LucidityDesign, line 113: Please set a name

LucidityDesign (changes requested):

src/components/content/Tooltip/Tooltip.stories.tsx
filiptammergard: Good catch, fixed now!

LucidityDesign (approved):

PR #910, opened 2023-01-16
fix(Alert): put overlay above sheet

filiptammergard:
To prevent weird situations like this:
image


LucidityDesign (commented):
I can't test it but the code looks good. Should the alert be above tooltips as well? Is it possible to add tooltips to alerts?

filiptammergard:
> I can't test it but the code looks good. Should the alert be above tooltips as well? Is it possible to add tooltips to alerts?

I think tooltips should be on the highest level. Tooltips in alerts are possible: 
image

PR #918, opened 2023-01-18
Multi select

LucidityDesign:


filiptammergard:
Really nice progress! One thing I noticed is that you can click "between" two dropdown options and the dropdown will close. It should probably not be possible to click between two options at all. Do you see what I mean?

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,373 @@
...
+    label: <Paragraph>Ornare Egestas Ridiculus</Paragraph>,
+    value: "Ornare Egestas Ridiculus",
+    description: "description two",
+  },
+  {
+    key: "75da75b2-11c5-48b9-b008-1b0659289ea2",
+    label: <Paragraph>Ridiculus Elit Inceptos</Paragraph>,
+    value: "Ridiculus Elit Inceptos",
+    description: "description three",
+  },
```
filiptammergard, line 52: Do you think it would make sense to use something like Faker.js for this?

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,373 @@
...
+  optionProps: {
+    "data-testid": "options",
+  },
+  clearButtonProps: {
+    "data-testid": "clear-button",
+  },
+}
+Mouse.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const inputField = canvas.getByTestId("input-1")
```
filiptammergard, line 303: When possible, I try to avoid using test ids to get elements. Is it possible to query these using `getByRole` instead? This is a great explanation: https://testing-library.com/docs/queries/about/#priority

src/components/controls/selects/MultiSelect/MultiSelectInput.tsx
```diff
@@ -0,0 +1,417 @@
...
+  background-color: ${({ theme, isSelected }) =>
+    isSelected ? theme.colors.background.primaryInverted : theme.colors.background.tertiary};
+  ${({ theme, isSelected }) =>
+    isSelected ? `color: ${theme.colors.content.primaryInverted}` : ""};
+  border-radius: ${({ theme }) => theme.borderRadii.sm};
+  padding: ${({ theme }) => 0.5 * theme.spacingBase}rem ${({ theme }) => theme.spacingBase}rem;
+  white-space: nowrap;
+  overflow: hidden;
+
+  p {
```
filiptammergard, line 368: I think this is a little brittle, since if someone uses a span or just plain text in the pill, this won't take effect. What do you think?

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
LucidityDesign: Not sure as I need static data to test the search functionality. If I create the data in a loop it will be much cleaner

src/components/controls/selects/MultiSelect/MultiSelectInput.tsx
LucidityDesign: I was wondering if allowing JSX is a good idea here. It gives more flexibility if the dev wants to add icons etc. but can be abused easily

src/components/controls/selects/MultiSelect/MultiSelect.mocks.tsx
```diff
@@ -0,0 +1,36 @@
...
+  "Nullam quis risus",
+  "eget urna mollis",
+  "ornare vel eu leo",
+]
+
+export const getMockData = (count: number): BaseOption[] => {
+  return [...Array(count)].map((_, index) => {
+    const content = contents[index % contents.length]
+    return {
+      key: `${index}`,
```
filiptammergard, line 30: I wonder if `key` really is needed for options in `<MultiSelect>`. It was added in `<SearchSelect>` for historic reasons, but `value` should always be unique, right? It would be nice to remove `key` if it's not necessary.

src/components/controls/selects/MultiSelect/MultiSelect.mocks.tsx
```diff
@@ -0,0 +1,36 @@
+import { Paragraph } from "../../../../main"
```
filiptammergard, line 1: Let's import from the source file to avoid circular dependency problems in the future!

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+Basic.args = {
+  label: "Label",
+  options: basicOptions,
+}
+
+export const DontClearAfterSelect = Template.bind({})
+DontClearAfterSelect.args = {
+  ...Basic.args,
+  clearSearchAfterSelect: false,
+}
```
filiptammergard, line 31: It would be nice to add an interaction test for this functionality as well to avoid regressions 🙂

src/components/controls/selects/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,304 @@
...
+  /** Props passed to dropdown element. */
+  dropdownProps?: ComponentPropsWithoutRef<"div"> & { "data-testid": string }
+
+  /** Filtering function to be used to populate dropdown. Filters on `option.value` by default. */
+  filter?: (value: string, option: Option) => boolean
+
+  /** If `false`, consumer have control over which options to pass to dropdown. Defaults to `true`. */
+  isFilterable?: boolean
+
+  clearSearchAfterSelect?: boolean
```
filiptammergard, line 36: Some JSDocs would be great for this prop!

src/components/controls/selects/MultiSelect/MultiSelect.mocks.tsx
filiptammergard: Also, I wonder whether `inputValue` makes sense in this component as it does in `<SearchSelect>`. If it doesn't, perhaps we should define a separate interface for the `<MultiSelect>` options instead of reusing the interface from `<SearchSelect>`, since they are a bit different in use case.

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+export const WithoutLabel = Template.bind({})
+WithoutLabel.args = {
+  options: basicOptions,
+  "aria-label": "Label",
+}
```
filiptammergard, line 37: Is it possible to add short interaction test with just one simple assertion for all these stories, as a kind of smoke test?

src/components/controls/selects/MultiSelect/MultiSelectInput.tsx
```diff
@@ -0,0 +1,469 @@
...
+      )}
+      <Wrapper
+        role="combobox"
+        aria-haspopup
+        aria-expanded={isOpen}
+        aria-controls={`options-${id}`}
+        hasLabel={"label" in props}
+      >
+        <Scroller ref={scrollableRef}>
+          <ScrollContent style={{ flex: `0 0 ${contentInlineSize}px` }}>
```
filiptammergard, line 232: If would be nice to let `<ScrollContent>` extend `<Box>` here, so that you can make this instead:

```suggestion
          <ScrollContent flexBasis={`${contentInlineSize}px`}>
```

src/components/controls/selects/MultiSelect/MultiSelectInput.tsx
filiptammergard: Right now `<Pill>` is a `<button>`, which is an inline element. Inline elements only allow for inline children. So I think either `<Pill>` has to change element, or we should do something else to make sure an inline element is used. What do you think?

src/components/controls/selects/MultiSelect/MultiSelectInput.tsx
```diff
@@ -0,0 +1,469 @@
...
+    white-space: nowrap;
+    margin: 0;
+  }
+`
+
+const SelectedOptionsWrapper = styled.div<WrapperProps>`
+  display: flex;
+  flex: 0 1 auto;
+  gap: ${({ theme }) => theme.spacingBase}rem;
+  margin-inline-start: ${({ theme, hasLabel }) => (hasLabel ? -1 * theme.spacingBase : 0)}rem;
```
filiptammergard, line 432: Pretty sure you can do something like this if you don't want to set `margin-inline-start` when there is no label:

```suggestion
  margin-inline-start: ${({ theme, hasLabel }) => hasLabel && `${theme.spacingBase}rem`};
```

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
LucidityDesign: What do you think of that solution?

src/components/controls/selects/MultiSelect/MultiSelectInput.tsx
LucidityDesign: This would create a lot of css classes while adding content to the input field:
https://emotion.sh/docs/best-practices#use-the-style-prop-for-dynamic-styles

src/components/controls/selects/MultiSelect/MultiSelect.mocks.tsx
```diff
@@ -0,0 +1,35 @@
...
+  "eget urna mollis",
+  "ornare vel eu leo",
+]
+
+export const getMockData = (count: number): BaseOption[] => {
+  return [...Array(count)].map((_, index) => {
+    const content = contents[index % contents.length]
+    return {
+      label: <Paragraph>{content}</Paragraph>,
+      value: content.toLowerCase().replace(/\s/g, "-"),
```
filiptammergard, line 31: Is `.toLowerCase().replace(/\s/g, "-")` for showing off the difference between `label` and `value`? The strings are unique from the start, so this is not purely necessary right?

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+Basic.args = {
+  label: "Label",
+  options: basicOptions,
+}
+Basic.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const label = canvas.getByText("Label")
+  const inputField = canvas.getByRole("textbox")
+  await userEvent.click(label)
+  await expect(inputField).toHaveFocus()
```
filiptammergard, line 31: What do you think about only asserting the accessible name here? I got a little confused when the story clicked the input field for me, so I think perhaps should leave that for the Keyboard and Mouse stories. What do you think?

src/components/controls/selects/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+}
+
+export const DontClearAfterSelect = Template.bind({})
+DontClearAfterSelect.args = {
+  ...Basic.args,
+  clearSearchAfterSelect: false,
+}
+DontClearAfterSelect.play = async ({ canvasElement }) => {
```
filiptammergard, line 39: Maybe we can put this story below next to Keyboard and Mouse to have the true interactive stories at the bottom, while keeping the basic ones (Basic, Without Label, Controlled, Message, Error Message) at the top?

src/components/controls/selects/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,311 @@
...
+  /** Props passed to dropdown element. */
+  dropdownProps?: ComponentPropsWithoutRef<"div"> & { "data-testid": string }
+
+  /** Filtering function to be used to populate dropdown. Filters on `option.value` by default. */
+  filter?: (value: string, option: Option) => boolean
+
+  /** If `false`, consumer have control over which options to pass to dropdown. Defaults to `true`. */
+  isFilterable?: boolean
+
+  /** Set if search term should be cleared after selecting an option. */
```
filiptammergard, line 41: Maybe something like this? I think it's nice to communicate the default when there is one. And "set" might be referring to both setting to true or false.

```suggestion
  /** If true, search term will be cleared after selecting an option. Default is `true`. */
```

filiptammergard (commented):
Maybe this will come after Rui's state & interaction review, but when I select three options like this:

image

And then using my arrows, I can't see what I'm focusing on. I.e., selected and focus+selected state are the same.

Should that be fixed?

src/components/controls/selects/MultiSelect/MultiSelect.mocks.tsx
LucidityDesign: Yes, it's to better distinguish the two values/their purpose

src/components/controls/selects/MultiSelect/MultiSelectInput.tsx
```diff
@@ -0,0 +1,508 @@
...
+
+export interface MultiSelectInputSharedProps<Option> {
+  /** Input placeholder. */
+  placeholder?: string
+
+  /** Default is `neutral`. */
+  status?: Status
+
+  /** Message shown below input field. Can be used together with `status` to show a success or error message. */
+  message?: ReactNode
```
filiptammergard, line 59: It's not needed in the first release, but we should remember to add `messageProps` here as well!

filiptammergard (approved):
Don't forget to squash commits 🙂

PR #936, opened 2023-01-24
fix(Menu): map isOpen correctly to the Radix UI open prop

filiptammergard:
Discovered that `isOpen` doesn't do anything, since Radix UI takes `open`. 

Can't do plain `open={isOpen}`, since `open` is typed as optional `boolean` and doesn't accept `undefined`.

### DIFF
--- a/src/components/menus/Menu/Menu.tsx
+++ b/src/components/menus/Menu/Menu.tsx
@@ -15,6 +15,13 @@ interface MenuProps {
   onOpenChange?: (open: boolean) => void
 }
 
-export const Menu = ({ children, ...props }: MenuProps): JSX.Element => {
-  return <DropdownMenu.Root {...props}>{children}</DropdownMenu.Root>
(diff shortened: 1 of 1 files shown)


LucidityDesign (commented):
What do think of using the radix interface?
```
type MenuProps = DropdownMenu.DropdownMenuProps & {
  /** @deprecated The controlled open state of the dropdown menu. Must be used in conjunction with onOpenChange. */
  isOpen?: boolean
}
```

filiptammergard:
> What do think of using the radix interface?
> 
> ```
> type MenuProps = DropdownMenu.DropdownMenuProps & {
>   /** @deprecated The controlled open state of the dropdown menu. Must be used in conjunction with onOpenChange. */
>   isOpen?: boolean
> }
> ```

That removes all JSDocs... For example this:

image

Becomes this:

image

So I think we want to keep doing them ourselves!

LucidityDesign (approved):

PR #942, opened 2023-01-26
Search select

LucidityDesign:
- document types
- limit dropdown size, enable scrolling
- use mantine scroll hook
- add tests
- a11y

### DIFF
--- a/src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
+++ b/src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
@@ -1,7 +1,32 @@
+import { faker } from "@faker-js/faker"
+import { expect } from "@storybook/jest"
 import { ComponentMeta, ComponentStory } from "@storybook/react"
+import { within, userEvent } from "@storybook/testing-library"
 import { useState } from "react"
-import { Paragraph } from "../../../typography/Paragraph/Paragraph"
+import { Box } from "../../../layout/Box/Box"
 import { SearchSelect } from "./SearchSelect"
+import { BaseOption } from "./types"
+
+const mockContents = [
+  "Snowfall guzzler drapery",
+  "Remorse strike tartly",
+  "Operator dazzling breeding",
+  "Egestas Lorem Ullamcorper",
+  "Ornare Egestas Ridiculus",
+  "Ridiculus Elit Inceptos",
+]
+
+function getMockData(count: number, withInputValue?: boolean): BaseOption[] {
+  return [...Array(count)].map((_, index) => {
+    const content =
+      mockContents[index] ?? faker.random.words(faker.datatype.number({ min: 1, max: 6 }))
+    return {
+      label: <Box>{content}</Box>,
+      value: withInputValue ? faker.datatype.uuid() : content,
+      ...(withInputValue ? { inputValue: content } : {}),
+    }
+  })
+}
 
 export default {
   title: "Controls/Selects/SearchSelect",
@@ -13,26 +38,8 @@ export default {
   },
 } as ComponentMeta<typeof SearchSelect>
 
-const basicOptions = [
-  {
-    key: "option-1",
-    label: <Paragraph>Snowfall guzzler drapery</Paragraph>,
-    value: "Snowfall guzzler drapery",
-    description: "description one",
-  },
-  {
-    key: "option-2",
-    label: <Paragraph>Remorse strike tartly</Paragraph>,
-    value: "Remorse strike tartly",
-    description: "description two",
-  },
-  {
-    key: "option-3",
-    label: <Paragraph>Operator dazzling breeding</Paragraph>,
-    value: "Operator dazzling breeding",
-    description: "description three",
-  },
-]
+const basicOptions = getMockData(3)
+const largeDataset = getMockData(32)
 
 const Template: ComponentStory<typeof SearchSelect<(typeof basicOptions)[0]>> = (args) => {
   const [searchTerm, setSearchTerm] = useState("")
@@ -46,33 +53,32 @@ Basic.args = {
   label: "Label",
   options: basicOptions,
 }
+Basic.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  await expect(canvas.getByText("Label")).toBeInTheDocument()
+  await expect(canvas.getByRole("textbox")).toBeInTheDocument()
+}
+
+export const LargeDataset = Template.bind({})
+LargeDataset.args = {
+  label: "Label",
+  options: largeDataset,
+}
+LargeDataset.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  await expect(canvas.getByRole("textbox")).toBeInTheDocument()
+}
 
 export const WithoutLabel = Template.bind({})
 WithoutLabel.args = {
   options: basicOptions,
 }
+WithoutLabel.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  await expect(canvas.getByRole("textbox")).toBeInTheDocument()
+}
 
-const inputValueOptions = [
-  {
-    label: <Paragraph>Snowfall guzzler drapery</Paragraph>,
-    inputValue: "Snowfall guzzler drapery",
-    value: "snowfall-guzzler-drapery",
-    description: "description one",
-  },
-  {
-    label: <Paragraph>Remorse strike tartly</Paragraph>,
-    inputValue: "Remorse strike tartly",
-    value: "remorse-strike-tartly",
-    description: "description two",
-  },
-  {
-    label: <Paragraph>Operator dazzling breeding</Paragraph>,
-    inputValue: "Operator dazzling breeding",
-    value: "operator-dazzling-breeding",
-    description: "description three",
-  },
-]
-
+const inputValueOptions = getMockData(3, true)
 const InputValueTemplate: ComponentStory<typeof SearchSelect<(typeof inputValueOptions)[0]>> = (
   args,
 ) => {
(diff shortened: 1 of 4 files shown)


src/components/controls/selects/SearchSelect/types.ts
```diff
@@ -1,8 +1,12 @@
 import { ReactNode } from "react"
 
 export interface BaseOption {
+  /** @deprecated since 7.12 */
```
filiptammergard, line 4: What do you think about something like this? I think it's good to always provide a "way out" when deprecating something.

```suggestion
  /** @deprecated since v7.12.0. `key` is not needed when using `value`, since value have to be unique. Use `value` instead! */
```

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
```diff
@@ -46,13 +32,51 @@ Basic.args = {
...
+  options: largeDataset,
+}
 
 export const WithoutLabel = Template.bind({})
 WithoutLabel.args = {
   options: basicOptions,
 }
 
-const inputValueOptions = [
+const inputValueOptions = getMockData(3, false)
```
filiptammergard, line 54: Shouldn't this be the other way around? Passing `false` makes it not use `inputValue`, right?

src/components/controls/selects/SearchSelect/SearchSelect.mocks.tsx
```diff
@@ -0,0 +1,35 @@
...
+  "eget urna mollis",
+  "ornare vel eu leo",
+]
+
+export const getMockData = (count: number, withInputValue = true): BaseOption[] => {
+  return [...Array(count)].map((_, index) => {
+    const content = contents[index % contents.length]
+    return {
+      label: <Paragraph>{content}</Paragraph>,
+      value: `${content.toLowerCase().replace(/\s/g, "-")}_${index}`,
```
filiptammergard, line 31: Is appending `index` needed here? `content` is already unique, right?

filiptammergard (commented):
Looks really nice! Some of my comments on `<MultiSelect>` apply here as well, so maybe we can resolve them there first to know what to do here as well?

src/components/controls/selects/SearchSelect/SearchSelect.mocks.tsx
LucidityDesign: When I create more items than `contents.length`, content isn't unique anymore

package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {
     "@einride/hooks": "1.7.4",
     "@emotion/is-prop-valid": "1.2.0",
+    "@faker-js/faker": "7.6.0",
```
filiptammergard, line 25: @faker-js/faker is already in main branch, so rebasing should remove the need for this 🙂

src/components/controls/inputs/BaseInput/BaseInput.tsx
```diff
@@ -41,7 +41,7 @@ export const BaseInput = forwardRef<HTMLInputElement, BaseInputProps>(
     return (
       <Box {...wrapperProps}>
         {label && (
-          <StyledLabel {...labelProps} htmlFor={id}>
+          <StyledLabel {...labelProps} htmlFor={id} id={`label_${id}`}>
```
filiptammergard, line 44: Why is `id` needed for this label? If it's only needed for `<SearchSelect>`, perhaps we can pass it from there instead of adding it to all components?

src/hooks/useScrollIntoView.ts
```diff
@@ -0,0 +1,77 @@
...
+  axis: "x" | "y"
+  duration: number
+  offset: number
+  cancelable: boolean
+  isList: boolean
+}
+
+/**
+ * Use mantine scroll-hook to scroll elements into viewport without scrolling body.
+ */
```
filiptammergard, line 24: It would be nice to know how this differs in functionality from Mantine's hooks, since it's pretty complex. Is Mantine's hook unable to prevent scrolling body?

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
```diff
@@ -46,13 +31,57 @@ Basic.args = {
...
+}
+
+export const LargeDataset = Template.bind({})
+LargeDataset.args = {
+  label: "Label",
+  options: largeDataset,
+}
+LargeDataset.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const inputField = canvas.getByRole("textbox") as HTMLInputElement
```
filiptammergard, line 49: We don't need to type assert here, right?

```suggestion
  const inputField = canvas.getByRole("textbox")
```

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
```diff
@@ -46,13 +31,57 @@ Basic.args = {
...
+  await expect(inputField).toBeTruthy()
```
filiptammergard, line 50: Nitpick: Maybe makes more sense to use `.toBeInTheDocument()` here, since it's a little bit strange to think about an input element to be truthy?

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
```diff
@@ -82,16 +111,58 @@ const InputValueTemplate: ComponentStory<typeof SearchSelect<(typeof inputValueO
...
     option.value.toLowerCase().trim().includes(value.toLowerCase().trim()) ||
     option.description.toLowerCase().trim().includes(value.toLowerCase().trim()),
 }
+CustomFilter.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const inputField = canvas.getByRole("textbox") as HTMLInputElement
+  inputField.focus()
+  await userEvent.type(inputField, "DeScRiptiOn tHrEe", { delay: 10 })
+  await userEvent.keyboard("[Enter]")
+  await expect(inputField.value).toBe("Operator dazzling breeding")
```
filiptammergard, line 128: You can you `.toHaveValue()` here to avoid needed the type assertion.

```suggestion
  await expect(inputField).toHaveValue("Operator dazzling breeding")
```

src/components/controls/selects/SearchSelect/SearchSelect.tsx
```diff
@@ -169,7 +189,7 @@ export const SearchSelect = <Option extends BaseOption>({
   }
 
   return (
-    <Wrapper {...wrapperProps}>
+    <Wrapper role="combobox" aria-haspopup="listbox" {...wrapperProps} aria-expanded={isOpen}>
```
filiptammergard, line 192: Props to the left of `{...wrapperProps}` can be overridden by consumers while the ones on the right cannot, so it's good to be conscious here. Perhaps `role` and `aria-haspopup` shouldn't allow overrides?

src/components/controls/selects/SearchSelect/SearchSelect.tsx
```diff
@@ -185,17 +205,27 @@ export const SearchSelect = <Option extends BaseOption>({
         ref={inputRef}
       />
       {isOpen && !!filteredOptions && filteredOptions.length > 0 && (
-        <OptionsWrapper {...dropdownProps}>
+        <OptionsWrapper
+          role="listbox"
+          aria-labelledby={`label_${inputRef.current?.id}`}
```
filiptammergard, line 210: Is this what they are doing in Mantine for aria-labelling options?

filiptammergard (commented):
Nice work!

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
LucidityDesign: I think this is what happens internally if you call `getByRole`:
```
await expect(canvas.getByText("Label"))
```
So when doing it like that we save a duplicated test

src/components/controls/selects/SearchSelect/SearchSelect.tsx
LucidityDesign: Not sure if they ever have to change these attributes but I wanted to give some flexibility to the devs.

src/components/controls/selects/SearchSelect/SearchSelect.tsx
LucidityDesign: Yes

src/components/controls/inputs/BaseInput/BaseInput.tsx
LucidityDesign: Good point!

src/hooks/useScrollIntoView.ts
LucidityDesign: Tried to improve the docs

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
filiptammergard: Hmm not sure what you mean – despite what `getByRole` does internally, why would asserting an input element with `toBeTruthy` be preferrable to `toBeInTheDocument`? 🙂

LucidityDesign: I mean that `getByRole` already fails if the element is not present :)
So I can write the above in one line without adding another test

package.json
filiptammergard: Now the dependency is in both `dependencies` and `devDependencies`, should only be in `devDependencies` 🙂

src/components/controls/selects/SearchSelect/SearchSelect.tsx
filiptammergard: Ah that sounds good! Seems like Testing Library inlines it like this in their internal tests:

https://github.com/testing-library/dom-testing-library/blob/0ce0c7054dfa64d1cd65053790246aed151bda9d/src/__tests__/ariaAttributes.js#L166

Personally I'd prefer being explicit with the assertion over not having any `expect` at all (even though it might test the same thing twice in practice) because it makes to easy to see the intention with the test. But you choose!

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
filiptammergard: Ah that sounds good! Seems like Testing Library inlines it like this in their internal tests:

https://github.com/testing-library/dom-testing-library/blob/0ce0c7054dfa64d1cd65053790246aed151bda9d/src/__tests__/ariaAttributes.js#L166

Personally I'd prefer being explicit with the assertion over not having any `expect` at all (even though it might test the same thing twice in practice) because it makes to easy to see the intention with the test. But you choose!

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
```diff
@@ -82,16 +106,58 @@ const InputValueTemplate: ComponentStory<typeof SearchSelect<(typeof inputValueO
...
   ...Basic.args,
+  options: descriptionOptions,
   filter: (value, option) =>
     option.value.toLowerCase().trim().includes(value.toLowerCase().trim()) ||
     option.description.toLowerCase().trim().includes(value.toLowerCase().trim()),
 }
+CustomFilter.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const inputField = canvas.getByRole("textbox")
+  inputField.focus()
```
filiptammergard, line 120: In the spirit of making the tests resemble how users use the software as much as possible, I think a click/tab to focus is preferable over programmatic focus. But fine either way!

```suggestion
  await userEvent.click(label)
```

filiptammergard (approved):
When `@faker-js/faker` is only in `devDependencies` I think this is ready to go, if you're also ready 🙂

PR #943, opened 2023-01-26
chore: add table view template

filiptammergard:
Skeleton from adding templates. Putting `Templates` in top level of Storybook hierarchy.

Adding a custom navbar for now, that can be improved on and reused if we need it in many templates.

There are a few open questions around the template design, as for example changing typography size on mobile, search button in navbar, highlighted table row, pagination, removed columns on mobile etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "dependencies": {
     "@einride/hooks": "1.7.4",
     "@emotion/is-prop-valid": "1.2.0",
+    "@faker-js/faker": "7.6.0",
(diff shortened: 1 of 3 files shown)


LucidityDesign:
Love it! Is the idea that we create a new stories file for each template or export a new component?

LucidityDesign (approved):

filiptammergard:
> Love it! Is the idea that we create a new stories file for each template or export a new component?

I think a separate story file (in the templates directory) for each template would be nice 🙂

PR #958, opened 2023-01-30
fix(DatePicker): decrease transition duration

LucidityDesign:


filiptammergard (approved):
Should probably be a `fix` or a `feat`, right?

PR #959, opened 2023-01-30
feat(datePickers): set date format via property

LucidityDesign:


src/components/controls/dates/DateRangePicker/DateRangePicker.tsx
```diff
@@ -43,6 +43,9 @@ interface DateRangePickerBaseProps {
 
   /** Props passed to root element. */
   wrapperProps?: BoxProps
+
+  /** dayjs input format */
```
filiptammergard, line 47: What do you think about noting the default here as well?

```suggestion
  /** `dayjs` input format. Default is `YYYY-MM-DD`. */
```

filiptammergard (approved):
Should be a `feat:`, since it adds the feature of being able to customize format, right?

When comments are resolved, this can be merged!

PR #960, opened 2023-01-30
chore: make navbar reusable in templates

filiptammergard:


LucidityDesign:
What do you think of adding a story to a group called "mocks" or "wireframe" or something like that?

filiptammergard:
> What do you think of adding a story to a group called "mocks" or "wireframe" or something like that?

Do you mean for the navbar (and not the table view)?

LucidityDesign:
> > What do you think of adding a story to a group called "mocks" or "wireframe" or something like that?
> 
> Do you mean for the navbar (and not the table view)?

Yes, the navbar

filiptammergard:
Maybe something like this?

src/templates/components/Navbar/Navbar.stories.tsx
```diff
@@ -0,0 +1,18 @@
+import { ComponentMeta, Story as StoryType } from "@storybook/react"
+import { Box } from "../../../components/layout/Box/Box"
+import { Navbar } from "./Navbar"
+
+export default {
+  title: "Templates/Wireframes/Navbar",
+  component: Navbar,
+  decorators: [
+    (Story: StoryType) => (
+      // Resetting margin from global decorator. Should be remove the global decorator, or can it be used conditionally?
```
LucidityDesign, line 10: Maybe you can pass an argument to the story wrapper like that?
https://dev.to/tmikeschu/controlling-global-decorators-via-args-in-storybook-41lf

filiptammergard: Hmm, not sure how you mean. Maybe you can push a commit to show?

LucidityDesign: Please see 5b1cb72

filiptammergard:
Nice! Updated now to use this way in both places!

LucidityDesign (approved):

PR #962, opened 2023-01-30
feat(Tooltip): add close delay duration

LucidityDesign:


filiptammergard:
It seems like a good approach!

It introduces some weirdness though:

https://user-images.githubusercontent.com/44197016/215494759-65f84ff4-4273-467d-8a81-7c858c6bb27f.mov

src/components/content/Tooltip/Tooltip.tsx
```diff
@@ -86,11 +94,35 @@ const StyledTooltipTrigger = styled(RadixTooltip.Trigger, {
...
+    }
+  }
+  @keyframes fadeOut {
+    from {
+      opacity: 1;
+    }
+    to {
+      opacity: 0;
+    }
+  }
```
filiptammergard, line 127: What do you think about using the Emotion approach to keyframes?

src/components/content/Tooltip/Tooltip.tsx
```diff
@@ -86,11 +94,35 @@ const StyledTooltipTrigger = styled(RadixTooltip.Trigger, {
...
   color: ${({ theme }) => theme.colors.content.primaryInverted};
   background: ${({ theme }) => theme.colors.background.primaryInverted};
   border-radius: ${({ theme }) => theme.borderRadii.sm};
   padding-block: ${({ theme }) => 0.5 * theme.spacingBase}rem;
   padding-inline: ${({ theme }) => theme.spacingBase}rem;
   z-index: ${zIndex.tooltip};
+  animation: fadeIn 0.1s ease-out;
+
+  &[data-state="closed"] {
+    animation: ${({ closeDelayDuration }) => `fadeOut 0.1s ${closeDelayDuration}ms ease-in`};
```
filiptammergard, line 109: What do you think about using the theme properties here?

```tsx
${({ theme }) => theme.transitions.morph.duration};
```

filiptammergard:
I also stumbled upon the case where the tooltip does not close at all:
https://user-images.githubusercontent.com/44197016/215728961-8de82fe7-dc67-4da0-a0ee-e84b2367db39.mov

filiptammergard:
Let's mark this as "won't do" for now as it adds a bunch on small bugs. We can revisit it this becomes an important requirement by some team!

PR #969, opened 2023-01-31
feat(Banner): add banner component

LucidityDesign:


src/components/content/Banner/Banner.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { BackgroundColor, ContentColor } from "../../../lib/theme/types"
+import { useColorScheme } from "../../../main"
+import { Link, LinkProps } from "../../typography/Link/Link"
+
+type Status = "success" | "warning" | "fail" | "neutral"
+
+interface BannerProps {
+  status?: Status
+
+  label: string
```
filiptammergard, line 13: I wonder if `title` is a better name for this than `label`? Also, should probably allow `ReactNode` to allow icons and similar things?

Maybe we can do as in `<Alert>`, to allow both `title` and `children`, where `title` is used when you want to do it in the structured way, and `children` can be used when you need to do some custom things?

src/components/content/Banner/Banner.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  linkProps?: LinkProps<"a">
+}
+
+export const Banner = forwardRef<HTMLDivElement, BannerProps>(
+  ({ status = "neutral", label, linkProps }, ref) => {
```
filiptammergard, line 19: Is the default for `status` here needed?

src/components/content/Banner/Banner.tsx
```diff
@@ -0,0 +1,85 @@
...
+        ref={ref}
+      >
+        {label}
+        {linkProps && <StyledLink {...linkProps} />}
+      </StyledBanner>
+    )
+  },
+)
+
+const StyledBanner = styled.div<{ backgroundColor: BackgroundColor; color: ContentColor }>`
```
filiptammergard, line 34: What do you think about extending `<Box>` or `<Group>` here?

src/components/typography/Link/Link.tsx
```diff
@@ -45,7 +45,7 @@ const StyledAnchor = styled("a", {
   text-decoration: underline;
 
   &:hover {
-    color: ${({ textColor, theme }) => getHoverColor(textColor, theme)};
+    color: ${({ textColor, theme }) => textColor && getHoverColor(textColor, theme)};
```
filiptammergard, line 48: What did this fix?

src/components/typography/Link/Link.tsx
```diff
@@ -4,7 +4,7 @@ import { ComponentPropsWithoutRef, ElementType, ReactNode } from "react"
 import { getColor, getFont } from "../../../lib/theme/prop-system"
 import { ContentColor, Font, Theme } from "../../../lib/theme/types"
 
-type LinkProps<C extends ElementType> = {
+export type LinkProps<C extends ElementType> = {
```
filiptammergard, line 7: This is not really needed anymore, right?

src/components/content/Banner/Banner.tsx
```diff
@@ -0,0 +1,73 @@
...
+  children?: ReactNode
+}
+
+export const Banner = forwardRef<HTMLDivElement, BannerProps>(
+  ({ status, title, children }, ref) => {
+    const { colorScheme } = useColorScheme()
+    return (
+      <StyledBanner
+        backgroundColor={getMessageBackgroundColor(status)}
+        color={getMessageColor(status, colorScheme)}
```
filiptammergard, line 27: I think the easiest fix here is to just call this `textColor` instead of `color`. What do you think about that?

src/components/typography/Link/Link.tsx
LucidityDesign: The link has a default hover color even though no color is set. But as I have removed the link from banner I will reset this

filiptammergard (approved):
Nice! 🚢

PR #971, opened 2023-02-01
feat(Option): make option component available

filiptammergard:
- Renamed from `<MenuItem>` to `<Option>`, since we already expose a `<MenuItem>` component. I also think it might be a better name, since it's not only going to be used in menus.
- Added variants and selected state according to Figma component.
- Used `<Option>` instead of `<SearchSelectOption>` in `<SearchSelect>` component.


src/components/menus/Option/Option.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  /** Variant of the option. Default is `primary`. */
+  variant?: Variant
+}
+
+export const Option = forwardRef<HTMLDivElement, OptionProps>(
+  ({ children, icon, selected, variant = "primary", ...props }, forwardedRef) => {
+    return (
+      <Wrapper
+        data-selected={selected}
```
LucidityDesign, line 25: Should we set `aria-selected` as well or is this its parents responsibility?

filiptammergard: Added! We need to be a bit careful in how this is used and probably improve aria when we know more. Since this is just an option and we don't know the contexts in which it will be used, we can't set any `role` for example.

LucidityDesign (approved):

src/components/menus/Option/Option.tsx
LucidityDesign: Same with the `tabindex`. Right now there is no option that is actually focusable

filiptammergard: I added `tabIndex={0}` so at least it's focusable by tabbing. But agreed, where this is used, it should probably most often be used with `as="button"`.

PR #972, opened 2023-02-01
fix(Option): update according to new state designs

filiptammergard:
There were some design changes here! Also important when to use elevated colors and when not to use it.

### DIFF
--- a/src/components/controls/selects/SearchSelect/SearchSelect.tsx
+++ b/src/components/controls/selects/SearchSelect/SearchSelect.tsx
@@ -230,10 +230,9 @@ export const SearchSelect = <Option extends BaseOption>({
         <OptionsWrapper role="listbox" aria-labelledby={id} {...dropdownProps} ref={scrollableRef}>
           {filteredOptions?.map((option, index) => (
             <Option
+              focused={index === selectedIndex}
               key={option.key ?? option.value}
               variant="secondary"
-              selected={index === selectedIndex}
-              aria-selected={index === selectedIndex}
               onClick={(e) => {
                 e.stopPropagation()
                 handleOptionSelect(option)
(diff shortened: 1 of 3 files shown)


src/components/controls/selects/SearchSelect/SearchSelect.tsx
```diff
@@ -232,8 +232,6 @@ export const SearchSelect = <Option extends BaseOption>({
             <Option
               key={option.key ?? option.value}
               variant="secondary"
-              selected={index === selectedIndex}
-              aria-selected={index === selectedIndex}
```
filiptammergard, line 236: I removed this, since this state is not only used for when an option is actually selected, but also for when hovering an option. So I don't think it makes sense to specify these—if we do, we should have separate hover and selected states.

LucidityDesign: It's difficult to define a clear state: In Search- and MultiSelect the option is kind of selected (focused) when navigating with arrow keys. But that's another discussion

LucidityDesign: That removes the "focus" state if using the up/down arrows. Maybe we have to split it into two states "focused" and selected/checked

LucidityDesign (changes requested):

src/components/controls/selects/SearchSelect/SearchSelect.tsx
LucidityDesign: That removes the "focus" state if using the up/down arrows. Maybe we have to split it into two states "focused" and selected/checked

LucidityDesign (approved):

PR #973, opened 2023-02-01
feat: style clear button focus state

LucidityDesign:
- style SearchInput clear button
- style SearchSelectInput clear button
- fix (un)controlled state of SearchInput
- focus SearchInput after clear


filiptammergard:
I could be mistaken, but it doesn't seem to be completely well-aligned vertically? 
image

filiptammergard (approved):
👏

Would be nice to make sure it's aligned!

PR #977, opened 2023-02-02
feat(DatePicker): allow free input of date

LucidityDesign:


src/components/controls/dates/DatePicker/DatePicker.tsx
```diff
@@ -46,6 +46,9 @@ interface DatePickerBaseProps {
 
   /** `dayjs` input format. Default is `YYYY-MM-DD`. */
   inputFormat?: string
+
+  /** Allow free input of date */
```
filiptammergard, line 50: Let's be explicit about the default here?

```suggestion
  /** Whether to allow free input. Default is `false`. */
```

src/components/controls/dates/DatePicker/DatePicker.stories.tsx
```diff
@@ -92,6 +92,20 @@ ErrorMessage.args = {
...
+  placeholder: "Placeholder",
+  allowFreeInput: true,
+}
+AllowFreeInput.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement)
+  const input = canvas.getByRole("textbox", { name: "Label" })
+  await userEvent.type(input, "2023-03-06", { delay: 10 })
+  await userEvent.keyboard("[Enter]")
+  await expect(input).toHaveValue("2023-03-06")
+}
```
filiptammergard, line 107: This is a fun interaction test to look at! 😃

src/components/controls/dates/DatePicker/DatePicker.stories.tsx
```diff
@@ -92,6 +92,20 @@ ErrorMessage.args = {
   status: "fail",
 }
 
+export const AllowFreeInput = Template.bind({})
+AllowFreeInput.args = {
+  label: "Label",
+  placeholder: "Placeholder",
+  allowFreeInput: true,
+}
```
filiptammergard, line 100: What about extending `WithLabel` here as we do in the other stories?

```suggestion
AllowFreeInput.args = {
  ...WithLabel.args,
  allowFreeInput: true,
}
```

filiptammergard (approved):
Nice! When comments are resolved this is good to go!

src/components/controls/dates/DatePicker/DatePicker.tsx
```diff
@@ -153,8 +156,11 @@ const StyledDatePicker = styled(MantineDatePicker, {
       color: ${({ theme }) => theme.colors.content.secondary};
     }
     &:focus {
-      outline: ${({ theme }) => theme.spacingBase / 8}rem solid
-        ${({ theme }) => theme.colors.border.selected};
+      outline: none;
+    }
+    &:focus-visible {
+      box-shadow: 0px 0px 0px ${({ theme }) => theme.spacingBase / 8}rem
+        ${({ theme }) => theme.colors.border.selected} inset;
```
filiptammergard, line 163: Ah, Safari doesn't respect `border-radius` when using outline? We might have some other usages of `outline` that needs updating

src/components/controls/dates/DatePicker/DatePicker.tsx
```diff
@@ -153,8 +156,11 @@ const StyledDatePicker = styled(MantineDatePicker, {
       color: ${({ theme }) => theme.colors.content.secondary};
     }
     &:focus {
-      outline: ${({ theme }) => theme.spacingBase / 8}rem solid
-        ${({ theme }) => theme.colors.border.selected};
+      outline: none;
+    }
+    &:focus-visible {
```
filiptammergard, line 161: `:focus` and `:focus-visible` is kind of the same when it comes to input fields, right?

filiptammergard (approved):

src/components/controls/dates/DatePicker/DatePicker.tsx
LucidityDesign: Yes, should I combine these?

src/components/controls/dates/DatePicker/DatePicker.tsx
LucidityDesign: Yes! We are not consistent on how to set the focus style:

1. `box-shadow: inset 0 0 0 0.0625rem ${({ theme }) => theme.colors.border.selected};`
2. `box-shadow: 0 0 0 1px ${({ theme }) => theme.colors.border.selected};`
3. `border: ${({ theme }) => theme.spacingBase / 8}rem solid
          ${({ theme }) => theme.colors.border.selected};`
4. `outline: ${({ theme }) => theme.spacingBase / 8}rem solid currentColor;`

src/components/controls/dates/DatePicker/DatePicker.tsx
filiptammergard: I think it would be nice to combine them in `:focus` to avoid communicating that they are different

src/components/controls/dates/DatePicker/DatePicker.tsx
filiptammergard: Added a task to go through these as part of the states and interactions initiative: https://www.notion.so/einride/Focus-styling-is-not-consistent-7abfa2ffb2e947ba94c394a3a169d34f

Option 3 makes the component reflow in certain situations so we should avoid that. Option 4 is easiest to read, but should be avoided due to the Safari bug (when going through them it would be good to add a comment on why we're using `box-shadow` instead of `outline` to avoid doing the same mistake again). I think option 1 is what should be used whenever applicable!

PR #985, opened 2023-02-02
feat(SearchSelect): don't select options on blur

LucidityDesign:


filiptammergard (approved):
Great! Maybe tomorrow look into if we can add an interaction test for this to make sure we don't regress!

PR #987, opened 2023-02-03
test(SearchSelect): check function calls

LucidityDesign:


src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
```diff
@@ -184,4 +197,7 @@ Keyboard.play = async ({ canvasElement }) => {
   await userEvent.keyboard("[Enter]")
   await expect(inputField).toHaveValue("")
   await expect(inputField).toHaveFocus()
+  inputField.blur()
```
filiptammergard, line 200: Let's keep this story close to user keyboard behavior and avoid programmatic blur:

```suggestion
  await userEvent.keyboard("[Tab]")
```

src/components/controls/selects/SearchSelect/SearchSelect.stories.tsx
```diff
@@ -184,4 +197,7 @@ Keyboard.play = async ({ canvasElement }) => {
...
+  await expect(onOptionSelect).toHaveBeenCalledTimes(1)
+  await expect(onClearClick).toHaveBeenCalledTimes(1)
```
filiptammergard, line 202: Nice! What do you think about adding assertions on this just before and after selecting an option and clicking the clear button as well? That would make it really obvious what we're trying to test 🙂

filiptammergard (approved):

PR #1006, opened 2023-02-09
chore(calendars): ensure snapshots are rendered consistent

LucidityDesign:


filiptammergard (approved):
It's not pretty, but if it solves the problem I think it's worth it. Another strategy I was thinking about would be overriding `Date` in some way to make Storybook think "today" is always the same date. But maybe that's not possible?

LucidityDesign:
> It's not pretty, but if it solves the problem I think it's worth it. Another strategy I was thinking about would be overriding `Date` in some way to make Storybook think "today" is always the same date. But maybe that's not possible?

I was not able to do it without unwanted side effects

PR #1011, opened 2023-02-09
chore: convert view stories to CSF3

filiptammergard:


src/components/views/Alert/Alert.stories.tsx
```diff
@@ -1,50 +1,27 @@
...
 import { expect } from "@storybook/jest"
-import { ComponentMeta, ComponentStory } from "@storybook/react"
+import { ComponentMeta, ComponentStoryObj } from "@storybook/react"
 import { userEvent, within } from "@storybook/testing-library"
 import { PrimaryButton } from "../../controls/buttons/PrimaryButton/PrimaryButton"
 import { Alert } from "./Alert"
 
 export default {
   title: "Views/Alert",
   component: Alert,
```
LucidityDesign, line 10: Do you think it's worth having a look into controls? Changing e.g. `title` and `description` would be nice

LucidityDesign (approved):

src/components/views/Alert/Alert.stories.tsx
filiptammergard: Would be nice! I think we can take a look at that in a separate PR. It's a bit cumbersome since `navigationTitle` and `children` both take `ReactNode`, which is not working very smoothly with Storybook controls

PR #1029, opened 2023-02-10
chore: trigger workflow manually

filiptammergard:


LucidityDesign:
I'll have a look into alternative triggers: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

PR #1043, opened 2023-02-15
feat(TimeInput): add component

LucidityDesign:


src/components/controls/inputs/TimeInput/useRangeTimeInput.ts
```diff
@@ -0,0 +1,56 @@
...
+  onChangeTo: (event: ChangeEvent<HTMLInputElement>) => void
+  maxFrom?: string | undefined
+  minTo?: string | undefined
+  statusFrom: Status
+  statusTo: Status
+}
+
+export type Status = "fail" | "neutral"
+
+export const useRangeTimeInput = (
```
filiptammergard, line 16: Maybe some JSDocs on this hook as well? Explaining what this hooks does and what happens when range is invalid etc.

src/components/controls/inputs/TimeInput/TimeInput.stories.tsx
```diff
@@ -0,0 +1,219 @@
...
+  },
+} satisfies Story
+
+const RangeTemplate = (): JSX.Element => {
+  const { valueFrom, onChangeFrom, valueTo, onChangeTo, minTo, maxFrom, statusFrom, statusTo } =
+    useRangeTimeInput()
+  return (
+    <HorizontalLayout>
+      <TimeInput
+        /* status={statusFrom} */
```
filiptammergard, line 103: Is this intended? Seems like `statusForm` is currently unused

filiptammergard (commented):
Just thinking out load: Do we need some time format prop? See https://www.notion.so/einride/Dates-and-times-in-Saga-cd3a770ce2374869938f8dc3c6459f48?pvs=4

Mantine uses `format: "12" | "24"`, but maybe that's not possible with the native `<input type="time" />`?

Should it be possible to choose between `:` and `.` as separators?

LucidityDesign:
> Just thinking out load: Do we need some time format prop? See https://www.notion.so/einride/Dates-and-times-in-Saga-cd3a770ce2374869938f8dc3c6459f48?pvs=4
> 
> Mantine uses `format: "12" | "24"`, but maybe that's not possible with the native `<input type="time" />`?
> 
> Should it be possible to choose between `:` and `.` as separators?

This is not possible with the native time input. It will choose 12/24h format based on the user preference

filiptammergard (approved):
Looks good! 👏

PR #1044, opened 2023-02-15
feat(colors): update warning and accent colors

filiptammergard:
Based on the recent updates in https://www.figma.com/file/zX0qOVghVidk8vWZdd2nwO/Primitives?node-id=1783%3A735&t=NnCPlbqWQxgWe16u-0.


LucidityDesign (approved):
Looks great! And the snapshots were created as well 👍 
Can you double check the snapshot, please?

PR #1065, opened 2023-02-21
Mapping features

LucidityDesign:


package.json
```diff
@@ -32,7 +32,8 @@
     "@radix-ui/react-tabs": "1.0.2",
     "@radix-ui/react-tooltip": "1.0.3",
     "dayjs": "1.11.7",
-    "lodash.merge": "4.6.2"
+    "lodash.merge": "4.6.2",
+    "react-map-gl": "7.0.21"
```
filiptammergard, line 36: This should probably be an optional peer dependency right? Otherwise all packages will get this, even the ones that don't use maps.

filiptammergard (approved):

PR #1067, opened 2023-02-21
fix(SearchSelect): allow messageProps prop

filiptammergard:


src/components/controls/selects/SearchSelect/SearchSelect.tsx
```diff
@@ -32,6 +32,9 @@ interface SearchSelectBaseProps<Option> extends ComponentPropsWithoutRef<"input"
   /** Message shown below input field. Can be used together with `status` to show a success or error message. */
   message?: ReactNode
 
+  /** Props passed to message element. */
+  messageProps?: Omit<ComponentPropsWithoutRef<"span"> & { "data-testid"?: string }, "color">
```
LucidityDesign, line 36: Why did you omit `color` explicitly? Would it make sense to omit `style`?

LucidityDesign: Ah I see that's how it was defined previously. However the message element is a `p`-tag and not a `span`

LucidityDesign: Nice!

PR #1074, opened 2023-02-22
feat(SearchSelect): filter for fallback data

LucidityDesign:


src/components/controls/selects/SearchSelect/filterOptions.ts
```diff
@@ -28,5 +28,10 @@ export const defaultFilter = <Option extends BaseOption>(
   value: string,
   option: Option,
 ): boolean => {
-  return option?.value?.toLowerCase().trim().includes(value.toLowerCase().trim())
```
filiptammergard, line 31: I think those `?`s were added to avoid errors when some props weren't yet defined. Is it fine anyway?

LucidityDesign: hmm yea let's add them again. Better safe than sorry.. But it shouldn't be possible that they are undefined

filiptammergard (approved):

PR #1096, opened 2023-03-01
feat(calendars): set static date to prevent changing snapshots

LucidityDesign:


src/components/controls/dates/Calendar/Calendar.tsx
```diff
@@ -18,6 +18,9 @@ interface CalendarBaseProps {
 
   /** Props passed to root element. */
   wrapperProps?: BoxProps
+
+  /** Initial date (month and year) to be displayed. */
```
filiptammergard, line 22: What do you think about this?

```suggestion
  /** Initial month displayed. */
```

filiptammergard (approved):
Nice!

src/components/controls/dates/Calendar/Calendar.tsx
filiptammergard: I think that makes sense because in this context you are not referring to month as in "January", but a specific month in time as in "January 2023". It's kind of the same for example in the calendar where you can switch between months, which also include a year implicitly. Makes sense?
image

PR #1105, opened 2023-03-06
feat(maps): add map tooltip

LucidityDesign:


src/components/mapping/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,54 @@
...
+      closeButton={false}
+      offset={15}
+    >
+      <Card>{label}</Card>
+    </StyledPopup>
+  )
+}
+
+const StyledPopup = styled(Popup)`
+  &.dark-popup .mapboxgl-popup-content {
```
filiptammergard, line 35: Why is styling only needed in dark mode? Wouldn't it be better to set colors in both modes explicitly?

src/components/mapping/types/types.ts
```diff
@@ -0,0 +1,11 @@
+export interface Coordinate {
+  lat: number
+  lng: number
+}
+
+export type IconMask = {
```
filiptammergard, line 6: Why is one of these an `interface` and the other a `type`? Does it make sense to default to `interface`?

package.json
```diff
@@ -32,7 +32,8 @@
     "@radix-ui/react-tabs": "1.0.2",
     "@radix-ui/react-tooltip": "1.0.4",
     "dayjs": "1.11.7",
-    "lodash.merge": "4.6.2"
+    "lodash.merge": "4.6.2",
+    "mapbox-gl": "2.13.0"
```
filiptammergard, line 36: This should be a dev + optional peer dependency

src/components/mapping/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,43 @@
+import { ComponentMeta, ComponentStory, ComponentStoryObj } from "@storybook/react"
+import { Map } from "react-map-gl"
+import { mapboxStyleLight } from ".."
+import { Text } from "../../typography/Text/Text"
+import { MapTooltip as Tooltip } from "./Tooltip"
+
+const reactAppMapboxAccessToken = "API_KEY"
```
filiptammergard, line 7: Does it make sense to get the map working in Storybook? I'm assuming it doesn't work without an API key

src/components/mapping/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useColorScheme } from "../../../contexts/ColorSchemeProvider"
+import { Card } from "../../cards/Card/Card"
+import { Coordinate } from "../types/types"
+
+interface TooltipProps {
+  label: JSX.Element | null
+  position: Coordinate
+}
+
+export const MapTooltip = ({ label, position }: TooltipProps): JSX.Element | null => {
```
filiptammergard, line 14: How come tooltips allow React components while markers don't?

filiptammergard (commented):
Right now white mode looks like this in Storybook:
image

While dark mode looks like this:
image

They don't seem equivalent?

src/components/mapping/Tooltip/Tooltip.tsx
LucidityDesign: `Popup` is a mapbox component. The icons are drawn on a deck.gl IconLayer which doesn't support HTML. As long as we stick to deck.gl we can only render SVGs

package.json
```diff
@@ -90,10 +91,14 @@
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
     "framer-motion": ">=6.2.8",
+    "mapbox-gl": "2.13.0",
```
filiptammergard, line 94: To allow bumping minor and patch versions:

```suggestion
    "mapbox-gl": "^2.13.0",
```

.env.example
```diff
@@ -0,0 +1 @@
+VITE_MAPBOX_TOKEN=XXX
```
filiptammergard, line 1: Add `"files.insertFinalNewline": true,` to your IDE settings 🙂 


```suggestion
VITE_MAPBOX_TOKEN=XXX

```

.gitignore
```diff
@@ -12,3 +12,5 @@ coverage
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+.env
```
filiptammergard, line 16: Vite has some conventions for env variables. Local env files are named `.env.local` while versioned env files are named `.env`. What do you think about following these conventions?

https://vitejs.dev/guide/env-and-mode.html

```suggestion
*.local

```

.github/workflows/chromatic.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           onlyChanged: true
         env:
           NODE_OPTIONS: "--max_old_space_size=4096"
+          VITE_MAPBOX_TOKEN: ${{ secrets.VITE_MAPBOX_TOKEN }}
```
filiptammergard, line 39: Is this also using token from Evolve?

README.md
```diff
@@ -27,6 +27,14 @@ export const App = (): JSX.Element => {
...
 ```
 
+## Run Storybook locally
+
+Config a mapbox api key and start storybook:
+
+- `cp .env.example .env`
+- add `VITE_MAPBOX_TOKEN`
+- `yarn run storybook`
+
```
filiptammergard, line 37: I've kept the README.md to only include information relevant for package consumers. Information regarding local development is kept in CONTRIBUTING.md—there is a Development section here: https://github.com/einride/ui/blob/main/CONTRIBUTING.md#development

.github/workflows/chromatic.yml
LucidityDesign: I haven't added a valid token to github secrets yet

src/components/mapping/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { Card } from "../../cards/Card/Card"
+import { Coordinate } from "../types/types"
+
+interface TooltipProps {
+  label: JSX.Element | null
+  position: Coordinate
+}
+
+export const MapTooltip = ({ label, position }: TooltipProps): JSX.Element | null => {
```
filiptammergard, line 13: Pretty sure `null` is a valid `JSX.Element`:
```suggestion
export const MapTooltip = ({ label, position }: TooltipProps): JSX.Element => {
```

src/components/mapping/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,49 @@
+import styled from "@emotion/styled"
+import { Popup } from "react-map-gl"
+import "mapbox-gl/dist/mapbox-gl.css"
+
+import { Card } from "../../cards/Card/Card"
+import { Coordinate } from "../types/types"
+
+interface TooltipProps {
+  label: JSX.Element | null
```
filiptammergard, line 9: Right?

```suggestion
  label: ReactNode
```

src/components/mapping/Tooltip/Tooltip.tsx
LucidityDesign: I get `Type 'null' is not assignable to type 'Element'.` 🤔

src/components/mapping/Tooltip/Tooltip.tsx
filiptammergard, line 1: To avoid mixing this up with our other `<Tooltip>`, maybe we can call this `<MapTooltip>` for clarity?

src/components/mapping/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import { ComponentMeta, ComponentStory, ComponentStoryObj } from "@storybook/react"
+import { Map } from "react-map-gl"
+import { mapboxStyleLight, mapboxStyleDark } from ".."
+import { useColorScheme } from "../../../contexts/ColorSchemeProvider"
+import { Text } from "../../typography/Text/Text"
+import { MapTooltip as Tooltip } from "./Tooltip"
```
filiptammergard, line 6: What do you think about keeping it as `MapTooltip` all way through?

src/components/mapping/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ReactNode } from "react"
+import { Popup } from "react-map-gl"
+import { Card } from "../../cards/Card/Card"
+import { Coordinate } from "../types/types"
+
+// eslint-disable-next-line import/extensions
+import "mapbox-gl/dist/mapbox-gl.css"
+
+interface TooltipProps {
+  label: ReactNode
```
filiptammergard, line 11: What do you think of passing tooltip content as `children` instead?

src/components/mapping/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,51 @@
+import styled from "@emotion/styled"
+import { ReactNode } from "react"
+import { Popup } from "react-map-gl"
+import { Card } from "../../cards/Card/Card"
+import { Coordinate } from "../types/types"
+
+// eslint-disable-next-line import/extensions
+import "mapbox-gl/dist/mapbox-gl.css"
+
+interface TooltipProps {
```
filiptammergard, line 10: Let's add some JSDocs!

src/components/mapping/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+type Story = ComponentStoryObj<typeof Tooltip>
+
+const Template: ComponentStory<typeof Tooltip> = (args): JSX.Element => {
+  const { colorScheme } = useColorScheme()
+  return (
+    <Map
+      style={{
+        height: "80vh",
+      }}
+      id="uiMap"
```
filiptammergard, line 25: Does this ID do something? Is it used in the apps? If it's not important but still needs to be provided I think we should make the casing more familiar:

```suggestion
      id="ui-map"
```

filiptammergard (commented):
Some small final comments!

filiptammergard:
Seems like you were right about `JSX.Element | null` as return type—`ReactNode` seems to produce a bunch of errors

filiptammergard (approved):

PR #1108, opened 2023-03-08
feat(Makers): add new marker icons

LucidityDesign:


src/components/mapping/types/types.ts
```diff
@@ -0,0 +1,22 @@
+export interface Coordinate {
```
filiptammergard, line 1: I think we can either put `types.ts` in paret directory or make one file per type in `types` dir. `types/types.ts` just an excessive dir level to me. What do you think?

src/components/mapping/Markers/Markers.ts
```diff
@@ -0,0 +1,331 @@
...
+export const iconHubDark: IconMask = {
+  url: hubDark,
+  ...size,
+}
+export const iconChargerDark: IconMask = {
+  url: chargerDark,
+  ...size,
+}
+
+export const getIcon = (
```
filiptammergard, line 274: Some thoughts that we can discuss:
- Isn't this the only thing that needs to be exported? Rather than both exporting this and also all the individual icons? I think we should only have one of the options.
- I think we should rename it to `getMapIcon` or something similar to not cause confusion with the `<Icon>` icons. 
- Would it make sense to make this a hook to be able to read `colorScheme` internally instead?
- If `getIcon` is intended to be used I think we should add tests and JSDocs for it, for example to explain what `state` refers to.

src/components/mapping/Markers/helpers.ts
```diff
@@ -0,0 +1,3 @@
+export function capitalizeFirstLetter(string: string): string {
+  return string.charAt(0).toUpperCase() + string.slice(1)
+}
```
filiptammergard, line 3: In our [style guide](https://einride.engineering/docs/tech-radar/frontend/techniques/einride-frontend-style-guide#file-naming), we have this bullet:

> Don't use file names like utils.ts as they easily become dumps for random functions. Prefer semantic names like sort.ts for a group of sorting utility functions or sortNames.ts if only one sorting utility is exposed.

I think this file could be renamed to `capitalizeFirstLetter.ts`, and perhaps include a couple of unit tests?

filiptammergard (commented):
Let's keep directory names kebab-case—`markers` instead of `Markers` if that makes sense to you. Is the extra dir level here needed at all? 🙂 

Let's reserve UpperCamelCase to React components—`markers.ts` instead of `Markers.ts` since it doesn't contain React components. Make sense?

src/components/mapping/markers/useMapMarker.ts
```diff
@@ -0,0 +1,151 @@
...
+export const useMapMarker = (): UseMarkerReturnType<MarkerName, MarkerState> => {
+  const { colorScheme } = useColorScheme()
+  const getMarker = useCallback(
+    (name: MarkerName, state: MarkerState = "default") => {
+      const markerSelector: MarkerSelector = `${name}${capitalizeFirstLetter(state)}`
+
+      return {
+        url: markers[colorScheme][markerSelector],
+        ...size,
+      }
```
filiptammergard, line 77: What do you think about this compared to using objects like this:

```ts
const markers = {
  light: {
    site: {
      default: siteLight,
      hover: siteHoverLight,
      selected: siteSelectedLight,
    },
  },
}
```

Then I think it could be selected without string manipulation:

```ts
url: markers[colorScheme][name][selector]
```

src/components/mapping/markers/Markers.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <>
+      <img src={icon.url} alt={name} />
+      <Text variant="bodySm">{[name, state || "default"].join(", ")}</Text>
+    </>
+  )
+}
+
+const Template = (): JSX.Element => {
+  return (
+    <VerticalLayout>
```
filiptammergard, line 39: Let's use `<Stack>` instead of `<VerticalLayout>`, which probably should be deprecated soon.

src/components/mapping/markers/Markers.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+      {MARKER_NAMES.map((name) => {
+        return STATES.map((state) => {
+          return (
+            <HorizontalLayout key={`${name}_${state}`} style={{ alignItems: "center" }}>
```
filiptammergard, line 43: Let's use `<Group>` instead of `<HorizontalLayout>`, which probably should be deprecated soon.

src/components/mapping/markers/Markers.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+              <Vehicle name={name} state={state} />
+            </HorizontalLayout>
+          )
+        })
+      })}
+    </VerticalLayout>
+  )
+}
```
filiptammergard, line 91: I think it's important to provide simple code examples in stories since people are often referring to them to understand how to do something. The abstractions for rendering all markers in one story make it a bit difficult to understand how to simply get one marker.

What do you think about doing as in `Icon.stories.tsx` – keep this story as `All` and also having a `One` story where you can choose what marker to render?

Instead of this story tree:

´´´
Mapping
|--Markers
|       |--Sites
 |      |--Vehicles
```

I think this would make sense:

```
Mapping
|--Sites
|       |--All
|       |--One
|--Vehicles
|       |--All
|       |--One

filiptammergard (commented):
Nice work! It looks good, have a few small suggestions/questions.

Is it intended that some of the icons have no circular border in light mode:
image


While they have one in dark mode?

image

src/components/mapping/index.ts
```diff
@@ -1,21 +1,4 @@
...
-import SiteHoveredIconDark from "./siteHoveredDark.svg"
-import SiteSelectedIcon from "./siteSelected.svg"
-import SiteSelectedIconDark from "./siteSelectedDark.svg"
-import SiteSmallIcon from "./siteSmall.svg"
-import SiteSmallIconDark from "./siteSmallDark.svg"
-import SiteWithChargerIcon from "./siteWithCharger.svg"
-import SiteWithChargerIconDark from "./siteWithChargerDark.svg"
-import SiteWithChargerSmall from "./siteWithChargerSmall.svg"
-import SiteWithChargerSmallDark from "./siteWithChargerSmallDark.svg"
+export * from "./types"
```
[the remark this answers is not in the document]
LucidityDesign: So far I need all exports, but I agree makes it easier to read and to modify

src/components/mapping/markers/capitalizeFirstLetter.ts
```diff
@@ -0,0 +1,3 @@
+export function capitalizeFirstLetter<T extends string>(string: T): Capitalize<T> {
+  return (string.charAt(0).toUpperCase() + string.slice(1)) as Capitalize<T>
+}
```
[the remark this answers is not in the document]
LucidityDesign: If I add a `MarkerName` I get a `Capitalize<MarkerName>` back. It's a bit more specific

PR #1125, opened 2023-03-13
feat(Menu): conditionally render menu content in portal

filiptammergard:


LucidityDesign:
There are two problems:
1.  the `z-index` is off. That could be solved by adding a `zIndex` property to override the default style
2. When rendering an opened menu (`defaultOpen={true}`) in a sheet, not all styles are set correctly. The menu dropdown still has `pointer-events: none` and the overlay has wrong styles as well. Once you close and open the context menu within the sheet, those issues are gone.

I think the second issue is not a realistic scenario as context menus shouldn't appear open within a sheet.

filiptammergard:
> There are two problems:
> 
> 1. the `z-index` is off. That could be solved by adding a `zIndex` property to override the default style
> 2. When rendering an opened menu (`defaultOpen={true}`) in a sheet, not all styles are set correctly. The menu dropdown still has `pointer-events: none` and the overlay has wrong styles as well. Once you close and open the context menu within the sheet, those issues are gone.
> 
> I think the second issue is not a realistic scenario as context menus shouldn't appear open within a sheet.
> 
> What do you think if [d1892b6](https://github.com/einride/ui/commit/d1892b65a25165cf4582a2e3e10c7fbcd28609d6)?

Yeah, this is probably a good idea. It'd be amazing to rely on something other than z-index, for example stacking contexts or layers, but that is probably much bigger changes and might not even be possible. I'll check this out deeper soon!

LucidityDesign (approved):

PR #1140, opened 2023-03-15
Revert "Revert "chore: turn main.js and preview.jsx into typescript""

LucidityDesign:
Reverts einride/ui#1137


filiptammergard (approved):
Nice feeling reverting the reverts!

PR #1146, opened 2023-03-16
chore: mirror file structure in storybook and code

filiptammergard:
This will base story hierarchy in Storybook on file structure in `src/components` instead of explicitly giving the hierarchy with `title` for each story. The main benefit is that code discovery will be easier since it follows the same structure as Storybook out of the box.

Made the change only for a few components as a start. If you agree, we can make it for all/most. It's still possible to define `title` explicitly for cases where it's needed.


LucidityDesign (approved):
Yes, that's great! 🚀

PR #1147, opened 2023-03-16
chore: add story introduction page

LucidityDesign:


filiptammergard (commented):
Nice!

src/stories/Introduction.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+### Content colors
+
+<ThemeColor colors={themes.light.colors.content} />
+
+## Dark Mode
+
+### Colors
+
+<ThemeColor colors={themes.dark.colors} />
```
filiptammergard, line 29: Since the doc pages doesn't support dark mode – can we wrap these components in dark mode so that they are rendered in the way they will be used in practice?

src/stories/Introduction.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+<Meta title="Introduction" />
+
+# Welcome to Einride UI
+
+Browse example stories now by navigating to them in the sidebar. View their code in the `stories`
+directory to learn how they work.
+
+## Light Mode
+
+### Colors
```
filiptammergard, line 13: Maybe a bit confusing with Colors, Background colors and Content colors—should we call them accent colors or something?

LucidityDesign: 👍 are they used?

filiptammergard: Not much I think. Let's talk with Rui about the reasoning behind these special colors—would be good to be able to give a short description to each category of colors. But this is good for now!

src/docs/ThemeColorWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+  align-items: center;
+`
+
+const Color = styled.div<{ background: string }>`
+  padding: 1rem;
+  border: 1px solid #aaa;
+  margin-inline-end: 1rem;
+  box-sizing: border-box;
+  width: 2rem;
+  height: 2rem;
```
filiptammergard, line 60: I think this should be aligned with the rest of our code style: Using theme for spacing and logical properties.

src/docs/ThemeColorWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+`
+
+const Color = styled.div<{ background: string }>`
+  padding: 1rem;
+  border: 1px solid #aaa;
+  margin-inline-end: 1rem;
+  box-sizing: border-box;
```
filiptammergard, line 58: This isn't needed as it is the default, right?

src/docs/ThemeColorWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+  flex-wrap: wrap;
+  background: ${({ background }) => background};
+  color: ${({ color }) => color};
+  padding: 1rem;
+`
+
+const ColorWrapper = styled.div`
+  margin-block-end: 0.5rem;
+  flex: 0 0 50%;
+  box-sizing: border-box;
```
filiptammergard, line 49: Not needed?

src/docs/ThemeColorWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+  display: flex;
+  align-items: center;
+`
+
+const Color = styled.div<{ background: string }>`
+  padding: 1rem;
+  border: 1px solid #aaa;
```
filiptammergard, line 56: Use border color from theme?

src/docs/ThemeColorWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+  margin-inline-end: 1rem;
+  box-sizing: border-box;
+  width: 2rem;
+  height: 2rem;
+  border-radius: 50%;
+  background: ${({ background }) => background};
+`
+
+const Text = styled.span`
+  user-select: all;
```
filiptammergard, line 66: Personal preference, but I think it's good to avoid messing with defaults for selecting text. This makes it impossible to select a part of the HEX for example. What do you think?

filiptammergard (approved):

PR #1149, opened 2023-03-16
chore: run storybook preview steps on PRs only

filiptammergard:
Actions fails since it tries to comment on PR when pushing to `main`: https://github.com/einride/ui/actions/runs/4434934246/jobs/7781458355


LucidityDesign (approved):
Thanks! LGTM

PR #1157, opened 2023-03-17
chore: add project specific style guide

filiptammergard:
Here are some of the styles we're currently kind of following and that I think we should follow going forward. Thoughts?

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,0 +1,195 @@
+# Style
+
+We're using
+[Einride Frontend Style Guide](https://einride.engineering/docs/tech-radar/frontend/techniques/einride-frontend-style-guide)
+as a base.
+
+In addition, here follows a few project specific styling conventions.
+
+## Components
+
+- Both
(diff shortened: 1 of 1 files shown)


STYLE.md
```diff
@@ -0,0 +1,191 @@
...
+    const button = canvas.getByRole("button", { name: Keyboard.args.children })
+    await expect(button).not.toHaveFocus()
+    await userEvent.tab()
+    await expect(button).toHaveFocus()
+  },
+} satisfies Story
+```
+
+- Every component should have a `Snapshot` story at the end of the file used for visual regression
+  testing. It ensures both color schemes are tested without extra configuration.
```
LucidityDesign, line 175: Maybe add something like this?
> Multiple variations/variants/stories are grouped together to decrease the [amount of snapshots](https://www.chromatic.com/pricing)

LucidityDesign (approved):
Great!

PR #1171, opened 2023-03-20
fix(dates): don't forward mantine props

LucidityDesign:


src/components/controls/dates/DatePicker/DatePicker.tsx
```diff
@@ -114,7 +115,7 @@ interface StyledDatePickerProps {
 }
 
 const StyledDatePickerInput = styled(DatePickerInput, {
-  shouldForwardProp: (prop) => prop !== "hasLabel", // avoid passing `hasLabel` attribute to HTML element
+  shouldForwardProp: (prop) => isPropValid(prop) && prop !== "withCellSpacing", // avoid passing `hasLabel` attribute to HTML element
```
filiptammergard, line 118: Should be enough right? Since `hasLabel` and `willCellSpacing` are not valid HTML attributes?

```suggestion
  shouldForwardProp: (prop) => isPropValid(prop),
```

src/components/controls/dates/DateRangePicker/DateRangePicker.tsx
```diff
@@ -115,7 +116,7 @@ interface StyledDateRangePickerProps {
 }
 
 const StyledDatePickerInput = styled(DatePickerInput, {
-  shouldForwardProp: (prop) => prop !== "hasLabel", // avoid passing `hasLabel` attribute to HTML element
+  shouldForwardProp: (prop) => isPropValid(prop) && prop !== "hasLabel", // avoid passing `hasLabel` attribute to HTML element
```
filiptammergard, line 119: Should be the same:

```suggestion
  shouldForwardProp: (prop) => isPropValid(prop),
```

filiptammergard (approved):
Let's make this a `fix` commit since it's not a new feature. After that, ship away

LucidityDesign:
This stops forwarding the props to the Mantine component 🤔

PR #1172, opened 2023-03-20
chore: rename mouse stories to poiner

filiptammergard:
Pointer better encapsulates what we're testing.


LucidityDesign (approved):
Please fix typo in commit and ship

PR #1183, opened 2023-03-21
fix: trim whitespace

LucidityDesign:


filiptammergard (approved):
Let's make this a `fix` to trigger a patch release instead of a minor release!

PR #1184, opened 2023-03-21
fix: define matrix of workflows

LucidityDesign:


LucidityDesign:
I had to add a workaround to comply with the name based branch protection rules

filiptammergard (approved):
We can try this, but I wonder if it's not just easier to add a separate step for each package, to avoid needing the branch protection hack?

LucidityDesign:
> We can try this, but I wonder if it's not just easier to add a separate step for each package, to avoid needing the branch protection hack?

We would need to add a new protection rule for each package that we add. But with 2, potentially 3 packages, it's not a big deal. What do you prefer?

PR #1194, opened 2023-03-23
feat: add map component

LucidityDesign:


filiptammergard:
Cool stuff! This work will probably be paused for a while now, but good to have this to look at in the future.

PR #1211, opened 2023-03-27
chore: separate einride-ui-maps to its own package

filiptammergard:


LucidityDesign:
Very cool!! Nice to see the final solution :)

filiptammergard:
@LucidityDesign Yeah! It's not working perfectly yet, but at least it's working 🙂

PR #2008, opened 2023-09-01
fix: add styling and position props to tooltip

KrisRosDim:
Added the option to style the tooltip content wrapper as well as change it's position.


stenehall (approved):
Nice improvements! The side attr was much needed.

PR #2070, opened 2023-09-20
fix: remove node version requirements

mblodau1337:
**Overview**
Remove the node version requirements from all packages.


filiptammergard (approved):
🙏

PR #2263, opened 2023-12-12
fix: forward ref to Menu component

Akari0731:
Forward ref to `Menu` component.
Allow wrapping `Menu` component with a component who requires to pass ref to child component like `Tooltip`.


mblodau1337 (approved):
🙌

PR #2291, opened 2024-01-08
chore: update mantine to latest

Akari0731:
It seems that mantine notification has an issue in book app when trying to update mantine to v7. I found that [@mantine/notifications v7.1.3 fixed some issue](https://github.com/mantinedev/mantine/issues/5007) and to make the debugging going forward, I'd like to bump all mantine packages to be the latest versions.

### DIFF
--- a/packages/einride-ui-dates/package.json
+++ b/packages/einride-ui-dates/package.json
@@ -6,8 +6,8 @@
     "build": "rollup --config"
   },
   "dependencies": {
-    "@mantine/core": "7.1.1",
-    "@mantine/dates": "7.1.1",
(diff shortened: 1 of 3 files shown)


filiptammergard:
Thanks @Akari0731, I'd really like to do this, but have been blocked by a change in Mantine types that causes the errors you see in CI here... 🥲

Akari0731:
Noooooooooooooo

filiptammergard:
I know 😢 I'll talk with Martin today to try to figure out what we can do about this

Akari0731:
[Update] Turns out I don't need this change right now for book-app, so feel free to close it:)

PR #2414, opened 2024-04-03
chore: introduce snapshots for storybook tests

Jassob:
This commit uses jest-image-snapshot to render the stories so that we
can catch dependency bumps where the UI changes inadvertently.


Jassob:
Apparently the animations are not deterministic, so I'll change the snapshot logic to be more selective.

PR #2582, opened 2024-10-22
chore: update node version to 22

Akari0731:
Since node v22 will become LTS in next week, we can safely recommend v22 for this repo. Setting engine configuration is just a recommendation so updating here for documentation purpose.
https://nodejs.org/en/about/previous-releases


yarn.lock
```diff
@@ -3692,29 +3692,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@radix-ui/react-dismissable-layer@npm:1.1.0":
```
Akari0731, line 3695: node 22 updated package installation a bit!

rollup.config.mjs
```diff
@@ -1,6 +1,6 @@
 import image from "@rollup/plugin-image"
 import typescript from "@rollup/plugin-typescript"
-import pkg from "./package.json" assert { type: "json" }
+import pkg from "./package.json" with { type: "json" }
```
Akari0731, line 3: node 22 require to use `with` instead of `assert`

